PR #10, opened 2015-01-06
Ticket 42: Clarify overly-redacted

eranmes:
And how TLS clients could check whether a domain name in a precertificate is overly redacted.


benlaurie:
LGTM

PR #12, opened 2015-01-06
Ticket 31: Clarify treatment of certificates without SCTs

eranmes:
As we do not desire to specify client behaviour, simply indicate that
certificates not accompanied by an SCT are not compliant (it's up to
the client to decide what to do with non-compliant certificates).


benlaurie:
LGTM

PR #14, opened 2015-01-06
Ticket 21: Specify cert acceptance criteria.

eranmes:
Require logs to accept certificates that meet a well-defined validity criteria.
This should address Steve Kent's concern about variation in logs' behaviour.
By requiring that logs accept certificates that are valid according to
X.509's validity requirement, a submitter can be certain their certificates
will be accepted.


benlaurie:
LGTM

PR #17, opened 2015-02-05
Clock skew

benlaurie:


benlaurie:
@eranmes

benlaurie:
@robstradling

robstradling:
LGTM

rfc6962-bis.xml
```diff
@@ -784,6 +784,12 @@ messages.
         In general, where needed, the <spanx style="verb">version</spanx> is v1 and the <spanx style="verb">id</spanx> is the log id for the log server queried.
       </t>
       <t>
+	In practice, log servers may be comprised of multiple machines. Since it is impractical to keep these machines in perfect sync, errors may occur that are caused by skew between the machines. Where such errors are possible, the log server will return additional information making it possible for clients to make progress, if progress is possible. Servers MUST only serve data that is free of gaps (that is, for example, no server will respond with an STH unless it is also able to prove consistency from all log entries logged within that STH).
```
eranmes: re ... the log servers will return additional information ...:
Is this optional? Can you also clarify how it should be done?

rfc6962-bis.xml
```diff
@@ -930,7 +936,7 @@ an accepted root certificate.
             </list></t>
           </list>
 
-          Both tree sizes must be from existing v1 STHs (Signed Tree Heads).
+          Both tree sizes must be from existing v1 STHs (Signed Tree Heads). However, because of skew, the receiving server may not know one or both of the existing STHs. If both are known, then only the <spanx style="verb">consistency</spanx> output is returned. If the first is known but not the second, then the latest known STH is returned, along with a consistency proof between the first STH and the latest.
```
eranmes: Presumably if both STHs are unknown consistency will be entirely omitted and only the latest known STH is returned?
Also, nit: How about using the term 'front-end' rather than 'server' to disambiguate the 'CT log server' entity from its components?

rfc6962-bis.xml
```diff
@@ -965,7 +983,7 @@ an accepted root certificate.
             </list></t>
 	  </list>
 
-          The <spanx style="verb">hash</spanx> must be calculated as defined in <xref target="tree"/>. The <spanx style="verb">tree_size</spanx> must designate an existing v1 STH.
+          The <spanx style="verb">hash</spanx> must be calculated as defined in <xref target="tree"/>. The <spanx style="verb">tree_size</spanx> must designate an existing v1 STH. Because of skew, the log server may not know the requested STH. In that case, it will return the latest STH it knows, along with an inclusion proof to that STH. If the server knows the requested STH then only <spanx style="verb">leaf_index</spanx> and <spanx style="verb">audit_path</spanx> are returned.
```
eranmes: Nit: replace "requested STH" with "requested tree size".
Also, how about the case when the front-end cannot provide an inclusion proof because the tree size it knows of does not contain the requested leaf? Will it then return the latest STH it knows about or nothing at all?

rfc6962-bis.xml
benlaurie: It is specified in the individual calls. I will say something to that effect.

rfc6962-bis.xml
benlaurie: Actually, I was planning to have an error in that case, but on second thoughts returning the STH makes more sense.

I have switched to front-end throughout.

rfc6962-bis.xml
benlaurie: We already say that the tree size specifies an STH, so it seems fine to say STH from then on?

benlaurie:
@eranmes PTAL

rfc6962-bis.xml
```diff
@@ -1045,6 +1077,9 @@ than the permitted number of entries, the log SHALL return the maximum number
 of entries permissible. These entries SHALL be sequential beginning with the
 entry specified by <spanx style="verb">start</spanx>.
         </t>
+	<t>
+	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
```
eranmes: Change log server to front-end here as well?

rfc6962-bis.xml
eranmes: Agreed - re-reading the paragraph it does make sense. How about changing log server to front-end here though?

rfc6962-bis.xml

eranmes:
LGTM

benlaurie:
d7291e1946d2f845f35f7468d109b347d5daa49f

PR #19, opened 2015-02-09
Add an "all-in-one" proof fetcher (trac ticket 19).

benlaurie:
@eranmes @robstradling

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1035,7 +1035,22 @@ an accepted root certificate.
 	  </list>
 	</t>
 	<t>
-          The <spanx style="verb">hash</spanx> must be calculated as defined in <xref target="tree"/>. The <spanx style="verb">tree_size</spanx> must designate an existing v1 STH. Because of skew, the front-end may not know the requested STH. In that case, it will return the latest STH it knows, along with an inclusion proof to that STH. If the front-end knows the requested STH and it is the latest STH then only <spanx style="verb">leaf_index</spanx> and <spanx style="verb">audit_path</spanx> are returned. If the front-end knows the requested STH and there is a new STH available, then it returns the new STH, an <spanx style="verb">audit_path</spanx> for that STH and a consistency proof for that STH with the requested STH (see <xref target="fetch_consistency"/>).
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1016,6 +1016,58 @@ an accepted root certificate.
...
+		A base64-encoded v1 leaf hash.
+	      </t>
+	      <t hangText="tree_size:">
+		The tree_size of the tree on which to base the proofs, in decimal.
+	      </t>
+            </list></t>
+	  </list>
+	</t>
+	<t>
+          The <spanx style="verb">hash</spanx> must be calculated as defined in <xref target="tree"/>. The <spanx style="verb">tree_size</spanx> must designate an existing v1 STH. Because of skew, the front-end may not know the requested STH. In that case, it will return the latest STH it knows, along with an inclusion proof to that STH. If the front-end knows the requested STH and it is the latest STH then only <spanx style="verb">leaf_index</spanx> and <spanx style="verb">audit_path</spanx> are returned. If the front-end knows the requested STH and there is a new STH available, then it returns the new STH, an <spanx style="verb">audit_path</spanx> for that STH and a consistency proof for that STH with the requested STH (see <xref target="fetch_consistency"/>).
```
eranmes: (1) What happens if the front-end knows of a (presumably smaller) tree that does not yet contain the leaf for which the inclusion proof is requested? 
(2) If the front-end knows of a smaller tree than the one designated by the client but includes the leaf the client is interested in, I suggest we make it explicit it MUST return the latest STH it knows of and an inclusion proof up to that STH (so the client can later audit the consistency of the STH it knows about and the STH served by the front-end).
(3) If a front-end knows of a tree size that's bigger than the requested tree_size I suggest it MUST return the latest STH it knows about, consistency proof between the client-designated STH AND inclusion proof to the latest STH (so the STH can only be dropped from the response if the STH designated by the tree_size in the request is precisely the latest one the front-end knows about).

eranmes:
Second version LGTM.

robstradling:
LGTM

benlaurie:
Hmm. I don't think the language leaves any doubt on concatenation. But the following sentence needs a MUST. Which I will add before pushing.

benlaurie:
Closed with e3ddb2a1da0b0551e0f83fcb27ed5a7d438a5fe1.

PR #22, opened 2015-02-17
Reveal the number of redacted labels.  Change (PRIVATE) to ?

robstradling:
http://trac.tools.ietf.org/wg/trans/trac/ticket/60

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -445,8 +445,8 @@ Validation Certificates</xref>.
         <section title="Redacting Domain Name Labels in Precertificates" anchor="redacting_subdomains">
           <t>
             When creating a Precertificate, the CA MAY substitute one or more
-labels in each DNS-ID with the label
-<spanx style="verb">?</spanx>. Every label to the left of a
+labels in each DNS-ID with a corresponding number of
+<spanx style="verb">?</spanx> labels. Every label to the left of a
 <spanx style="verb">?</spanx> label MUST also be a
 <spanx style="verb">?</spanx> label. For example, if a certificate contains a
 DNS-ID of <spanx style="verb">top.secret.example.com</spanx>, then the
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -444,12 +444,12 @@ Validation Certificates</xref>.
         </section>
         <section title="Redacting Domain Name Labels in Precertificates" anchor="redacting_subdomains">
           <t>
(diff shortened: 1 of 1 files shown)


robstradling:
@benlaurie @eranmes

rfc6962-bis.xml
```diff
@@ -444,12 +444,12 @@ Validation Certificates</xref>.
...
-the complete leftmost labels in each DNS-ID with the literal string
-<spanx style="verb">(PRIVATE)</spanx>. For example, if a certificate contains a
+            When creating a Precertificate, the CA MAY substitute each of the
+one or more leftmost labels in each DNS-ID with the label
+<spanx style="verb">?</spanx>. For example, if a certificate contains a
 DNS-ID of <spanx style="verb">top.secret.example.com</spanx>, then the
 corresponding Precertificate could contain
-<spanx style="verb">(PRIVATE).example.com</spanx> instead. Labels in a
+<spanx style="verb">?.?.example.com</spanx> instead. Labels in a
 <xref target="RFC6125">CN-ID</xref> MUST remain unredacted.
```
benlaurie: Possibly we should be a bit more generic here?

"Labels MUST NOT be redacted such that it becomes impossible to determine whether the certificate is correctly issued. For example, labels in a CN-ID MUST remained unredacted."

rfc6962-bis.xml
```diff
@@ -444,12 +444,12 @@ Validation Certificates</xref>.
         </section>
         <section title="Redacting Domain Name Labels in Precertificates" anchor="redacting_subdomains">
           <t>
-            When creating a Precertificate, the CA MAY substitute one or more of
-the complete leftmost labels in each DNS-ID with the literal string
-<spanx style="verb">(PRIVATE)</spanx>. For example, if a certificate contains a
+            When creating a Precertificate, the CA MAY substitute each of the
+one or more leftmost labels in each DNS-ID with the label
```
benlaurie: Hmm. Scans horribly.

How about "When creating a Precertificate, each label, starting from the left, may be replaced with a ?".

rfc6962-bis.xml
benlaurie: Oh, we already say it below - CN-ID stuff should perhaps move down?

robstradling: The CN-ID stuff will be changing (see ticket #17), but I was planning to tackle that in a separate pull request.

robstradling:
Merged at 82ee686fa39a4fc10b6bca05aca0aa3d6ca5afd3

PR #23, opened 2015-02-17
Use correct type for precert_entry_V2 in TimestampedEntry.signed_entry.

robstradling:
This bug must've crept in when we updated the Precertificate format.
"PreCert" is no longer defined or used anywhere else.


robstradling:
@benlaurie @eranmes

benlaurie:
LGTM

robstradling:
Merged at 8ecd62c81e256be941aba9fd6e8753bfacedee3a

PR #25, opened 2015-02-18
More redaction updates

robstradling:


robstradling:
@benlaurie @eranmes

eranmes:
LGTM

rfc6962-bis.xml
```diff
@@ -447,12 +447,22 @@ Validation Certificates</xref>.
...
-<spanx style="verb">?.?.example.com</spanx> instead. Labels in a
-<xref target="RFC6125">CN-ID</xref> MUST remain unredacted.
+<spanx style="verb">?</spanx> label MUST also be redacted. For example, if a
+certificate contains a DNS-ID of
+<spanx style="verb">top.secret.example.com</spanx>, then the corresponding
+Precertificate could contain <spanx style="verb">?.?.example.com</spanx>
+instead, but not <spanx style="verb">top.?.example.com</spanx> instead.
+          </t>
+          <t>
+            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted.
```
benlaurie: Hmm. Why not?

rfc6962-bis.xml
```diff
@@ -447,12 +447,22 @@ Validation Certificates</xref>.
...
+However, if the complete leftmost label of a DNS-ID is
+<spanx style="verb">*</spanx>, it is considered redacted for the purposes of
+determining if the label to the right may be redacted. For example, if a
+certificate contains a DNS-ID of
+<spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding
+Precertificate could contain
+<spanx style="verb">*.?.?.example.com</spanx> instead, but not
+<spanx style="verb">?.?.?.example.com</spanx> instead.
```
benlaurie: Don't get it. How would a client process these two differently?

rfc6962-bis.xml
robstradling: * labels aren't always permitted (e.g. they're forbidden in EV certs), so, to aid compliance checking (with, for example, the EV SSL Guidelines), it would be useful for CT to always reveal * labels.  I can't think of a legitimate need to redact a * label.

rfc6962-bis.xml
robstradling: A TLS client would see a cert containing *.top.secret.example.com and would discover (from the .2.4.6 certificate extension) that the 3 leftmost labels are redacted in the Precertificate.  The TLS client would know that * is considered redacted, so it would only do two substitutions - "top" for "*" and "secret" for "*" - when it reconstructs the TBSCertificate.
If the TBSCertificate component of the Precertificate contains *.?.?.example.com, then the SCT signature will verify correctly (unless there are other problems).
But if the TBSCertificate component of the Precertificate contains ?.?.?.example.com, the TLS client will consider the SCT signature to be invalid.
(What a TLS client does when it encounters an invalid SCT signature is out of scope for this pull request, I think).

benlaurie:
Makes sense. LGTM.

robstradling:
Merged as 49de5eb87e1e6d29e50c418c00654a79a950522d

PR #28, opened 2015-02-20
Add errors for machine processing.

benlaurie:
@eranmes.

BTW, I intend to submit a later CL that brings the list layout into line for other sections.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -879,7 +879,7 @@ messages.
 		</t>
               </list>
 	    </t>
-            <t hangText="Errors:">
+            <t hangText="Error codes:">
 	      <list style="hanging">
 		<t hangText="unknown root">
 		  The root of the chain is not one accepted by the log.
@@ -993,7 +993,7 @@ an accepted root certificate.
               </list>
               Note that no signature is required on this data, as it is used to verify an STH, which is signed.
             </t>
-	    <t hangText="Errors:">
+	    <t hangText="Error codes:">
 	      <list style="hanging">
 		<t hangText="first unknown">
 		  <spanx style="verb">first</spanx> is before the latest known STH but is not from an existing STH.
@@ -1046,7 +1046,7 @@ an accepted root certificate.
 		</t>
               </list>
 	    </t>
-	    <t hangText="Errors:">
+	    <t hangText="Error codes:">
 	      <list style="hanging">
 		<t hangText="hash unknown">
 		  <spanx style="verb">hash</spanx> is not the hash of a known leaf (may be caused by skew or by a known certificate not yet merged).
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -816,15 +816,19 @@ messages.
         If the log is unable to process a client's request, it MUST return an HTTP response code of 4xx/5xx (see <xref target="RFC2616"/>), and, in place of the responses outlined in the subsections below, the body SHOULD be a JSON structure containing at least the following field:
         <list style="hanging">
           <t hangText="error_message:">
-            <list>
-              <t>
                 A human-readable string describing the error which prevented the log from processing the request.
-              </t>
-              <t>
-                In the case of a malformed request, the string SHOULD provide sufficient detail for the error to be rectified.
-              </t>
-            </list>
           </t>
+          <t>
+                In the case of a malformed request, the string SHOULD provide sufficient detail for the error to be rectified.
+          </t>
+	  <t hangText="error_code:">
+	    An error code readable by the client. Some codes are generic and are detailed here. Others are detailed in the individual requests. Error codes are fixed text strings.
+            <list style="hanging">
+              <t hangText="not compliant">
+		The request is not compliant with this RFC.
+	      </t>
+	    </list>
+	  </t>
         </list>
       </t>
       <t>
@@ -835,6 +839,7 @@ messages.
           <artwork>
   {
     "error_message": "'start' cannot be greater than 'end'",
+    "error_code": "not compliant",
   }
           </artwork>
         </figure>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -846,46 +851,51 @@ messages.
...
+		</t>
+		<t hangText="extensions:">
+		  An opaque type for future expansion. It is likely that not all participants will need to understand data in this field. Logs should set this to the empty string. Clients should decode the base64-encoded data and include it in the SCT.
+		</t>
+		<t hangText="signature:">
+		  The SCT signature, base64 encoded.
+		</t>
+              </list>
+	    </t>
+            <t hangText="Errors:">
```
eranmes: lowercase e to match the rest of the response values?

rfc6962-bis.xml
```diff
@@ -946,47 +956,52 @@ an accepted root certificate.
...
+		<t hangText="sha256_root_hash:">
+		  The Merkle Tree Hash of the tree, in base64.
+		</t>
+		<t hangText="tree_head_signature:">
+		  A TreeHeadSignature for the above data.
+		</t>
+              </list>
+              Note that no signature is required on this data, as it is used to verify an STH, which is signed.
+            </t>
+	    <t hangText="Errors:">
```
eranmes: same - lowercase e

rfc6962-bis.xml
```diff
@@ -946,47 +956,52 @@ an accepted root certificate.
...
+	      <list style="hanging">
+		<t hangText="first unknown">
+		  <spanx style="verb">first</spanx> is before the latest known STH but is not from an existing STH.
+		</t>
+		<t hangText="second unknown">
+		  <spanx style="verb">second</spanx> is before the latest known STH but is not from an existing STH.
```
eranmes: Do you get the error *in addition* to the latest STH in case the second sth provided is unknown? the description of the method makes it clear than in case of a known first sth but unknown second sth the latest sth (+ consistency proof) is returned. no mentions of error.

rfc6962-bis.xml
benlaurie: This is not a response value, it is a title (like Inputs and Outputs).

rfc6962-bis.xml
```diff
@@ -846,46 +851,51 @@ messages.
...
+              </list>
+	    </t>
+            <t hangText="Errors:">
+	      <list style="hanging">
+		<t hangText="unknown root">
+		  The root of the chain is not one accepted by the log.
+		</t>
+		<t hangText="bad chain">
+		  The alleged chain is not actually a chain of certificates.
+		</t>
```
eranmes: While we're at it, can we please have "bad certificate" which says the certificate is not encoded using valid DER?

rfc6962-bis.xml
benlaurie: The error is "not from an existing STH" - i.e. a tree size that does not correspond to an issued STH.

benlaurie: Our log would probably never return this, because we'd provide a proof anyway, but logs are allowed to decline to provide proofs.

rfc6962-bis.xml
```diff
@@ -994,47 +1009,50 @@ an accepted root certificate.
...
+              </list>
+	    </t>
+	    <t hangText="Errors:">
+	      <list style="hanging">
+		<t hangText="hash unknown">
+		  <spanx style="verb">hash</spanx> is not the hash of a known leaf (may be caused by skew).
+		</t>
+		<t hangText="tree_size unknown">
+		  <spanx style="verb">hash</spanx> is before the latest known STH but is not from an existing STH.
```
eranmes: What's the error if the tree_size is after the latest known STH?

rfc6962-bis.xml
```diff
@@ -994,47 +1009,50 @@ an accepted root certificate.
...
+		</t>
+		<t hangText="tree_head_signature:">
+		  A TreeHeadSignature for the above data.
+		</t>
+              </list>
+	    </t>
+	    <t hangText="Errors:">
+	      <list style="hanging">
+		<t hangText="hash unknown">
+		  <spanx style="verb">hash</spanx> is not the hash of a known leaf (may be caused by skew).
```
eranmes: There's the case (which Al found out, actually) where the hash is not unknown, but the leaf has yet to be incorporated into the tree. Will this error be returned in this case?

rfc6962-bis.xml
benlaurie: There is no error - this is a normal condition and a non-error return is defined.

rfc6962-bis.xml
benlaurie: Yes. I will expand the parenthetical comment.

rfc6962-bis.xml
eranmes: Ah, each would be present independently, I see.
In that case, unknown_root and bad_chain to match 'sct_version' above?

rfc6962-bis.xml
eranmes: Sorry, should have been clearer:
If I send a get-sth-consistency with first designating an existing STH and second < tree_size but does not designate any STH the log produced, do I:
- Get an error?
- Get the latest STH + consistency proof from first to that?
- Get both?

IIUC you say that logs may return an error or may provide a consistency proof anyway, just want to be clear that the response shouldn't contain both.

rfc6962-bis.xml
benlaurie: sct_version is a field name, unknown root is a field value, so no obvious reason why they should match.

rfc6962-bis.xml
benlaurie: This is clear from the opening text (errors replace outputs).

rfc6962-bis.xml

eranmes:
@AlCutter and @pphaneuf for their opinion of the actual error codes.

benlaurie:
Pushed at 3ea4baf931c2b479340ab5916267949fc864c4c1.

PR #29, opened 2015-02-26
Make layout consistent.

benlaurie:
@eranmes

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -903,21 +903,21 @@ messages.
         <t>
           POST https://&lt;log server&gt;/ct/v1/add-pre-chain
         </t>
-        <t>
-          Inputs:
-
-          <list>
-            <t><list style="hanging">
-              <t hangText="precertificate:">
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -903,21 +903,21 @@ messages.
...
-          Inputs:
-
-          <list>
-            <t><list style="hanging">
-              <t hangText="precertificate:">
-                The base64-encoded Precertificate.
-              </t>
-              <t hangText="chain:">
-                An array of base64-encoded CA certificates. The first element is
+	<t>
```
eranmes: is this tag supposed to have different indentation than the one above it?

eranmes:
LGTM otherwise.

rfc6962-bis.xml
benlaurie: Hmm. This is tabs vs. spaces. Possibly we should axe tabs in another CL, coz I'm sure there are many!

PR #31, opened 2015-02-27
Ticket 39: Potential monitor behaviour.

eranmes:
Rather than specify exactly what a monitor could be doing in addition to log monitoring, give an example.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1320,7 +1320,7 @@ log's public key.
       </section>
       <section title="Monitor" anchor="monitor">
         <t>
-          Monitors watch logs and check that they behave correctly. They also watch for certificates of interest.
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1320,7 +1320,7 @@ log's public key.
       </section>
       <section title="Monitor" anchor="monitor">
         <t>
-          Monitors watch logs and check that they behave correctly. They also watch for certificates of interest.
+          Monitors watch logs and check that they behave correctly. Monitors may additionally watch for certificates of interest. For example, a monitor may be configured to report on all certificates that apply to a specific domain name, when fetching new entries for consistency validation.
```
benlaurie: I don't think the comma belongs there (s/name, when/name when/).

benlaurie:
LGTM.

rfc6962-bis.xml

PR #40, opened 2015-03-09
Intro is not informal, precertificate is not a proper noun.

benlaurie:


benlaurie:
@eranmes

benlaurie:
Fixes http://trac.tools.ietf.org/wg/trans/trac/ticket/62 and http://trac.tools.ietf.org/wg/trans/trac/ticket/62.

PR #41, opened 2015-03-09
Revamp section 3.4 and clarify ASN.1 encoding (ticket 14)

robstradling:
@eranmes @benlaurie

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -619,7 +619,7 @@ SignedCertificateTimestamp v2. Currently, no extensions are specified.
 from at least one log must be included in the TLS handshake by using one or more
 of the mechanisms listed below. Three mechanisms are provided because they have
 different tradeoffs. TLS clients MUST implement all three mechanisms. TLS
-servers MUST implement at least one of the three mechanisms.
+servers MUST present SCTs using at least one of the three mechanisms.
           <list style="symbols">
             <t>
               A TLS extension (Section 7.4.1.4 of <xref target='RFC5246'/>) with
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -613,44 +613,44 @@ TBSCertificate (in the case of a PrecertChainEntryV2), as described above.
 SignedCertificateTimestamp v2. Currently, no extensions are specified.
         </t>
       </section>
-
       <section title="Including the Signed Certificate Timestamp in the TLS Handshake">
         <t>
-The SCT data corresponding to at least one certificate in the chain from at
-least one log must be included in the TLS handshake, either by using
-an X509v3 certificate extension as described below, by using a TLS
-extension
-(Section 7.4.1.4 of <xref target='RFC5246'/>) 
-with type
-"signed_certificate_timestamp", or by using Online Certificate
-Status Protocol (OCSP) Stapling (also known as the "Certificate
-Status Request" TLS extension; see 
-<xref target='RFC6066'/>), where the OCSP response includes a non-critical
-extension with OID 1.3.6.1.4.1.11129.2.4.5 (see <xref target='RFC2560'/>) and
-body:
-
-
-        </t>
-        <figure>
-          <artwork>
-    SignedCertificateTimestampList ::= OCTET STRING</artwork></figure>
-        <t>
-          in the singleExtensions component of the SingleResponse pertaining to the end-entity certificate.
-        </t>
-        <t>
-          At least one SCT MUST be included. Server operators MAY include more than one SCT.
-        </t>
-        <t>
-          Similarly, a certification authority MAY submit a precertificate to more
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -613,44 +613,44 @@ TBSCertificate (in the case of a PrecertChainEntryV2), as described above.
...
+          <list style="symbols">
+            <t>
+              a TLS extension (Section 7.4.1.4 of <xref target='RFC5246'/>) with
+type <spanx style="verb">signed_certificate_timestamp</spanx>
+(see <xref target="tls_sctlist_extension"/>). This mechanism allows TLS servers
+to participate in CT without the cooperation of CAs, unlike the other two
+mechanisms. It also allows SCTs to be updated on the fly.
+            </t>
+            <t>
+              an <xref target='RFC6960'>Online Certificate Status Protocol
```
benlaurie: s/an/An/.

rfc6962-bis.xml
```diff
@@ -613,44 +613,44 @@ TBSCertificate (in the case of a PrecertChainEntryV2), as described above.
...
-receiving the certificate, clients can reconstruct the original TBSCertificate
-to verify the SCT signature.
+          The SCT data corresponding to at least one certificate in the chain
+from at least one log must be included in the TLS handshake by using one or more
+of the mechanisms listed below. Three mechanisms are provided because they have
+different tradeoffs. TLS clients MUST implement all three mechanisms. TLS
+servers MUST implement at least one of the three mechanisms.
+          <list style="symbols">
+            <t>
+              a TLS extension (Section 7.4.1.4 of <xref target='RFC5246'/>) with
```
benlaurie: s/a/A/

rfc6962-bis.xml
```diff
@@ -613,44 +613,44 @@ TBSCertificate (in the case of a PrecertChainEntryV2), as described above.
...
-certificate, by encoding the SignedCertificateTimestampList structure as an
-ASN.1 OCTET STRING and inserting the resulting data in the TBSCertificate as a
-non-critical X.509v3 certificate extension (OID 1.3.6.1.4.1.11129.2.4.2). Upon
-receiving the certificate, clients can reconstruct the original TBSCertificate
-to verify the SCT signature.
+          The SCT data corresponding to at least one certificate in the chain
+from at least one log must be included in the TLS handshake by using one or more
+of the mechanisms listed below. Three mechanisms are provided because they have
+different tradeoffs. TLS clients MUST implement all three mechanisms. TLS
+servers MUST implement at least one of the three mechanisms.
```
benlaurie: Slightly weird thing to say, since one of the mechanisms (embedded in cert) is not implemented by the server...

benlaurie: Perhaps something more like "TLS servers MUST present SCTs using at least on of the three..."

benlaurie:
In general, I've been using one long line per paragraph because reformatting otherwise tends to make diffs hard to review and not reformatting when there are line breaks tends to be ugly.

Not concerned enough to make a big thing of it, but wondering what your views are...

rfc6962-bis.xml
eranmes: +1 to Ben's suggestion - that reads much better.

robstradling: FWIW, that text is in the current draft, but I agree that Ben's suggestion reads much better.

robstradling:
It's a shame the draft currently contains a mixture of long lines and manual line breaks.  It would be nice to use the same style consistently.  I don't mind which, it's just that not using manual line breaks every <=80 characters would require me to break a deeply entrenched habit.  ;-)

benlaurie:
I agree it is a shame. Settling on one or the other would be good. Long lines probably work better with github (where I don't think you can ask for word diff).

rfc6962-bis.xml
```diff
@@ -658,8 +658,7 @@ to verify the SCT signature.
 
      struct {
          SerializedSCT sct_list &lt;1..2^16-1&gt;;
-     } SignedCertificateTimestampList;
-          </artwork>
+     } SignedCertificateTimestampList;</artwork>
```
benlaurie: </artwork> should be on its own line.

benlaurie: THis said the XML closing tag should be on its own line, but github ate it.

benlaurie:
LGTM.

rfc6962-bis.xml
robstradling: I deliberately moved the closing artwork tag to the end of the previous line, to get rid of a blank empty line at the bottom of the artwork (in the HTML output from http://xml2rfc.ietf.org/ at least).

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/f9b13d5333f3bd92c0cd247df32610f379074d2b

PR #42, opened 2015-05-18
Ticket 82: Adding the SCTs to the data returned by get-entries.

eranmes:
Note the version bump to get-entries.
@benlaurie @robstradling - one of you care to take a look?

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1176,7 +1176,7 @@ an accepted root certificate.
 
       <section title="Retrieve Entries from Log" anchor="fetch_entries">
         <t>
-          GET https://&lt;log server&gt;/ct/v1/get-entries
+          GET https://&lt;log server&gt;/ct/v2/get-entries
         </t>
         <t>
 	  <list style="hanging">
@@ -1200,6 +1200,9 @@ an accepted root certificate.
                     </t>
                     <t hangText="extra_data:">
                       The base64-encoded unsigned data pertaining to the log entry. In the case of an X509ChainEntry, this is the <spanx style="verb">certificate_chain</spanx>. In the case of a PrecertChainEntryV2, this is the whole <spanx style="verb">PrecertChainEntryV2</spanx>.
+                    </t>
+                    <t hangText="signed_certificate_timestamp:">
+                      A Signed Certificate Timestamp for the entry, in the same format as in <xref target="sct"/>.
                     </t>
 		  </list>
 		</t>


eranmes:
Would be of interest to @pphaneuf .

rfc6962-bis.xml
```diff
@@ -1201,6 +1201,9 @@ an accepted root certificate.
                     <t hangText="extra_data:">
                       The base64-encoded unsigned data pertaining to the log entry. In the case of an X509ChainEntry, this is the <spanx style="verb">certificate_chain</spanx>. In the case of a PrecertChainEntryV2, this is the whole <spanx style="verb">PrecertChainEntryV2</spanx>.
                     </t>
+                    <t hangText="signed_certificate_timestamp:">
+                      A Signed Certificate Timestamp for the entry, in the same format as in <xref target="sct"/>.
```
robstradling: add-(pre-)chain returns a broken-down SCT (i.e. each field as a separate parameter), whereas I think you're proposing here to put the whole SCT blob into "signed_certificate_timestamp".
How is this "the same format" ?

Note that "leaf_input" contains the base64-encoded MerkleTreeLeaf structure, which already contains the SCT's timestamp and extensions (in "timestamped_entry").

If we assume that the log ID is already known to the caller, then the only SCT fields missing from the existing get-entries specification are "sct_version" and "signature".
So how about adding "sct_version" and "signature" (although I'd call it "sct_signature" here), instead of "signed_certificate_timestamp" ?

benlaurie: As mentioned elsewhere, I suspect add-(pre-)chain should also return an SCT blob.

robstradling: OK.  In that case, "the same format" will become valid when we make that change.  :-)

Shall we also look at pruning the contents of "leaf_input" so that get-entries doesn't return any duplicated SCT fields?  I think we only need...
           LogEntryType entry_type;
           select(entry_type) {
               case x509_entry: ASN.1Cert;
               case precert_entry_V2: TBSCertificate;
           }

"LogEntryType entry_type" could become a field inside the SCT, as per TRAC ticket 10.

pphaneuf:
After discussing this with @eranmes yesterday, it did became clear to me that I wanted two different things: being able to get SCTs (provided here), and getting integrity validation on the `get-entries` responses.

I only filed one ticket because I thought getting the SCT would get my two birds with one stone, but @eranmes reminded me that the cert chain isn't covered by the signature in the SCT, so it wouldn't have integrity validation on that.

If I'm not mistaken, even though the RFC says that a log should support HTTPS, the protocol is self-standing, providing its own integrity and authentication, but `get-entries` is the odd exception, due to that "hole". It's important for the implementation of the distributed log to have integrity checking, and while we can implement it on the side by ourselves (providing a SHA256 of the body in an HTTP header, say), it seems like an odd blind spot?

benlaurie:
On 19 May 2015 at 13:28, Pierre Phaneuf <notifications@github.com> wrote:

> After discussing this with @eranmes <https://github.com/eranmes>
> yesterday, it did became clear to me that I wanted two different things:
> being able to get SCTs (provided here), and getting integrity validation on
> the get-entries responses.
>
> I only filed one ticket because I thought getting the SCT would get my two
> birds with one stone, but @eranmes <https://github.com/eranmes> reminded
> me that the cert chain isn't covered by the signature in the SCT, so it
> wouldn't have integrity validation on that.
>
> If I'm not mistaken, even though the RFC says that a log should support
> HTTPS, the protocol is self-standing, providing its own integrity and
> authentication, but get-entries is the odd exception, due to that "hole".
> It's important for the implementation of the distributed log to have
> integrity checking, and while we can implement it on the side by ourselves
> (providing a SHA256 of the body in an HTTP header, say), it seems like an
> odd blind spot?
>
Chains include their own integrity: each certificate is signed by the
certificate above. There might be a slight hole there, though - I can't
remember if we extend chains to include the root cert (the presented chain
is not required to include it). If not, then its possible you could not
know the appropriate root cert...

pphaneuf:
I see... But even if the root cert was included, it would leave that one unverified, right?

benlaurie:
On 19 May 2015 at 16:44, Pierre Phaneuf <notifications@github.com> wrote:

> I see... But even if the root cert was included, it would leave that one
> unverified, right?
>

Root certs are self-signed. So, no.

benlaurie:
This was superceded by #53.

PR #43, opened 2015-05-18
Adding log version as metadata.

eranmes:
To avoid forcing clients to probe themselves.


eranmes:
Manually committed as https://github.com/google/certificate-transparency-rfcs/commit/bf30d22bb8b5ef1f30c8abc229f571f60a3641b6

PR #45, opened 2015-06-08
Limit the number of STHs produced to avoid client marking attacks.

benlaurie:
@robstradling

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -791,9 +791,9 @@ signatures.
           The leaves of the Merkle Tree are the leaf hashes of the corresponding <spanx style="verb">MerkleTreeLeaf</spanx> structures.
         </t>
       </section>
-      <section title="Signed Tree Head">
+      <section title="Signed Tree Head" anchor="STH">
         <t>
-        Every time a log appends new entries to the tree, the log SHOULD sign the corresponding tree hash and tree information (see the corresponding Signed Tree Head client message in <xref target="fetch_sth"/>). The signature for that data is structured as follows:
+        Periodically the log SHOULD sign the corresponding tree hash and tree information (see the corresponding Signed Tree Head client message in <xref target="fetch_sth"/>). The signature for that data is structured as follows:
         </t>
 
 	<figure>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1295,6 +1295,12 @@ but it is expected there will be a variety.
 	  <t hangText="Maximum Merge Delay">
 	    The MMD the log has committed to.
 	  </t>
+	  <t hangText="STH Frequency Period">
+	    The time, in seconds, over which <spanx style="verb">STH Frequency Count</spanx> is measured (see <xref target="STH"/>).
+	  </t>
+	  <t hangText="STH Frequency Count">
+	    The maximum number of STHs the log may produce in any <spanx style="verb">STH Frequency Period</spanx> (see <xref target="STH"/>).
+	  </t>
```
robstradling: Wouldn't it be simpler to define just one field - "Minimum Merge Delay" - instead of these two STH Frequency fields?

mpalmer:
How will the contents of an STH be used to mark a client?  Isn't the only part of the STH that ever comes back to the server the number of entries?  If so, you can't get away with incrementing that without actually *adding* the entries, and a log which is adding entries quickly enough to be able to uniquely mark every client request is pretty quickly going to catch the attention of every monitor (if only because they'll have a tree size that is orders of magnitude larger than anyone else's).

rfc6962-bis.xml
benlaurie: I wanted to allow some slop, rather than insist on a hard minimum.

benlaurie:
@mpalmer This is true, but you still need to have some limit or you cannot objectively enforce a restriction.

rfc6962-bis.xml
robstradling: OK.

Does the "Misbehaving Logs" section (currently 8.4) need to deal with the scenario of a log issuing more than "STH Frequency Count" STHs in any given "STH Frequency Period" ?

eranmes: I also support adding less fields - what do you think about omitting  the STH Frequency Period so the STH Frequency Count relates to the MMD? (i.e. how many STHs the log may produce in a given MMD-sized time frame).

pphaneuf:
@mpalmer, the STH has a timestamp, a CT log could sign a new STH with the same tree size (and tree hash), but a different timestamp for every request, and remember who it handed each timestamp to. You'd then be able to who's sending STHs around...

benlaurie:
@pphaneuf Good point - that could be revealed by, e.g., gossip.

rfc6962-bis.xml
benlaurie: @eranmes I can go with that, or we can punt and just say "STH Frequency", since we don't actually define the metadata in detail.

benlaurie: @robstradling I don't think so - STH Frequency is a privacy, not security, consideration.

eranmes: @benlaurie I'm fine with either (I think that STH Count per MMD is more difficult to misinterpret, so easier to enforce when monitoring).

benlaurie:
@eranmes PTAL

ln5:
I think this is good.

Sadly enough, limiting the number of {tree head, tree size and timestamp} per time unit still leaves the attack described by @pphaneuf open because ECDSA signatures are typically non deterministic (i.e. changes for every signature mad over the same data). We'd need to mandate RFC6979 to stop that, see http://trac.tools.ietf.org/wg/trans/trac/ticket/83 .

I suppose that a mitigating action that could be taken by clients would be to "merge" STH:s with the same {head,size,timestamp} into one before sharing them through gossip.

eranmes:
@ln5 Good point about signatures being non-deterministic. Would you like to bring it up on the list?

benlaurie:
It is already in trac.

benlaurie:
Merged at 239e6a04ba1cbd0d5a9681d905a8104c98db874d.

PR #46, opened 2015-06-11
Remove unreferenced LogEntry structure

mpalmer:
The `LogEntryType` enum is used in a number of places, but the `LogEntry`
structure is never used anywhere.  I remember this causing me no end of
confusion when I was grappling with understanding RFC6962, and since the
structure hasn't seen any new use since, I'm thinking it can probably stand
to go.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -380,14 +380,6 @@ following components:
   <artwork>
     enum { x509_entry(0), precert_entry_V2(3), (65535) } LogEntryType;
 
-    struct {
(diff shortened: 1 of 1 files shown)


eranmes:
@benlaurie or @robstradling , see any issue with it? I'd assume the intention was to use the LogEntry structure in the digitally-signed part of the SignedCertificateTimestamp, but it never made it there.

benlaurie:
FWIW, the structure is actually used in our logs, but only internally.

PR #47, opened 2015-06-11
Make /get-entries?end= optional

mpalmer:
Given:

1. There is no way to know, in advance, the maximum number of entries which a log will return from a /get-entries call; and

2. Some logs will return a 400 if you ask for an `end` which is >= tree_size; and

2. It is inefficient to require a call to /get-sth to get the tree size before each request to /get-entries; and

4. I can think of no reason why the log shouldn't be able to easily introspect its size to determine where to end.

Therefore, be it resolved that I have submitted this Pull Request for your kind and valuable consideration.  <grin>

In practice, it makes writing a monitor that much easier if I don't have to keep track of where the end might be, and make sure I don't run off the end of it.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1179,7 +1179,7 @@ an accepted root certificate.
 
       <section title="Retrieve Entries from Log" anchor="fetch_entries">
         <t>
-          GET https://&lt;log server&gt;/ct/v1/get-entries
+          GET https://&lt;log server&gt;/ct/v2/get-entries
         </t>
         <t>
 	  <list style="hanging">
@@ -1189,7 +1189,7 @@ an accepted root certificate.
 		  0-based index of first entry to retrieve, in decimal.
 		</t>
 		<t hangText="end:">
-		  0-based index of last entry to retrieve, in decimal.
+		  (optional) 0-based index of last entry to retrieve, in decimal.  If this parameter is not specified, then the log should treat it as being equal to the <spanx style="verb">tree_size - 1</spanx>.
 		</t>
               </list>
 	    </t>


pphaneuf:
The only bit that seems like a problem to me is item 2 there. Not sure which logs do this, but the Google open source log server doesn't (it returns an error if `end` is smaller than `start`, which I think is fair), and I don't think the production logs (Aviator, Pilot, Rocketeer) do this? They just truncate to what they have?

It already says this:

> Logs MAY restrict the number of entries that can be retrieved per <spanx style="verb">get-entries</spanx> request.  If a client requests more than the permitted number of entries, the log SHALL return the maximum number of entries permissible.

Maybe it needs to be clarified to explicitly cover this case, but I feel that this means that if `end` goes past the tree size, it should return what it has, not give you an error.

mpalmer:
All of the Google logs appear to fall foul of item 2:

```
$ wget -O - https://ct.googleapis.com/pilot/ct/v1/get-entries?start=12345\&end=9999999999999
--2015-06-11 23:48:26--  https://ct.googleapis.com/pilot/ct/v1/get-entries?start=12345&end=9999999999999
Resolving ct.googleapis.com (ct.googleapis.com)... 2404:6800:4006:800::200a, 216.58.220.138
Connecting to ct.googleapis.com (ct.googleapis.com)|2404:6800:4006:800::200a|:443... connected.
HTTP request sent, awaiting response... 400 Bad Request
2015-06-11 23:48:28 ERROR 400: Bad Request.

$ wget -O - https://ct.googleapis.com/aviator/ct/v1/get-entries?start=12345\&end=9999999999999
--2015-06-11 23:48:37--  https://ct.googleapis.com/aviator/ct/v1/get-entries?start=12345&end=9999999999999
Resolving ct.googleapis.com (ct.googleapis.com)... 2404:6800:4006:800::200a, 216.58.220.138
Connecting to ct.googleapis.com (ct.googleapis.com)|2404:6800:4006:800::200a|:443... connected.
HTTP request sent, awaiting response... 400 Bad Request
2015-06-11 23:48:38 ERROR 400: Bad Request.

$ wget -O - https://ct.googleapis.com/rocketeer/ct/v1/get-entries?start=12345\&end=9999999999999
--2015-06-11 23:48:45--  https://ct.googleapis.com/rocketeer/ct/v1/get-entries?start=12345&end=9999999999999
Resolving ct.googleapis.com (ct.googleapis.com)... 2404:6800:4006:800::200a, 216.58.220.138
Connecting to ct.googleapis.com (ct.googleapis.com)|2404:6800:4006:800::200a|:443... connected.
HTTP request sent, awaiting response... 400 Bad Request
2015-06-11 23:48:47 ERROR 400: Bad Request.
```

Even if they didn't, though, setting `end` to an arbitrarily large number rather than just leaving it off seems... inelegant.

pphaneuf:
Ah, I just scolded @AlCutter for this. :smiley:

His argument is you'll usually have a `get-sth` call before a series of `get-entries` calls, and you don't need to call `get-sth` before *every* call to `get-entries`, just once in a while is sufficient. So it should be fairly trivial for a monitor to start off with a `get-sth`, and possibly not issue any `get-entries` at all, if it finds that it already has all the entries?

In the log monitoring code I wrote, I use specific `end` numbers that are not arbitrarily large, because I issue concurrent requests (so something like `start=0&end=999`, `start=1000&end=1999`, and so on). But I still also happens to know the end of the range I'm trying to fetch, so it wasn't any harder to have the last request be something like `start=2000&end=2437`...

robstradling:
@pphaneuf Al's right.  My monitor (for https://crt.sh) calls each log's get-sth once in a while (currently every 10 minutes).  After each get-sth call, it calls get-entries zero or more times to fetch all of the entries it hasn't yet seen.  No need to call get-sth every time.

eranmes:
I'm not in favour of changing the treatment of the end parameter. The RFC already allows the behaviour, but mandating it does not seem to provide such a significant benefit.
A monitor is expected to check for tree consistency and to do that it must fetch an STH. This change would only be useful for monitors which inspect the entries, not checking the tree consistency (and unless they check tree consistency, how could they know they receive all entries?).
In summary, this Pull Request proposes reducing the freedom in implementation of logs for all implementations for the benefit of a particular type of monitors,  so despite the thoughtful, concisely laid-out arguments, I don't think it should be merged. Sorry!

pphaneuf:
There's potential inconsistency, in the case where a log has multiple frontends: you could do your `get-sth` request on a frontend that is up-to-date, and then do the appropriate `get-entries` (with `end` being the tree size in the STH minus 1, as it should), but end up on a frontend that's slightly behind. One could argue that getting a 400 in that case is a bit rude?

But if the frontend returns what it has, you're still a bit stuck: you have an STH for a tree of size N, but only X entries (where X < N), what now? You can't verify the hash, so you retry the `get-entries` for the missing bit until it works?

If we were to take the same approach that this problem has been tackled elsewhere, I could imagine some sort of `get-entries-and-sth` call?

But in the case that @mpalmer is bringing up here, I just don't see it as much of an imposition to have to get an STH before getting entries... And it feels unsafe to me to have a default that could throw large amounts of data at you. If I'd set a default for `end`, I'd make it "the same as `start`" (and I think not setting a default is fine, so I'm not in favour of merging this pull request either).

mpalmer:
OK, seems like the consensus is against this.  No worries.

benlaurie:
On 11 June 2015 at 15:53, Pierre Phaneuf <notifications@github.com> wrote:

> There's potential inconsistency, in the case where a log has multiple
> frontends: you could do your get-sth request on a frontend that is
> up-to-date, and then do the appropriate get-entries (with end being the
> tree size in the STH minus 1, as it should), but end up on a frontend
> that's slightly behind. One could argue that getting a 400 in that case is
> a bit rude?
>

It absolutely is and should be fixed.


> But if the frontend returns what it has, you're still a bit stuck: you
> have an STH for a tree of size N, but only X entries (where X < N), what
> now? You can't verify the hash, so you retry the get-entries for the
> missing bit until it works?
>
> If we were to take the same approach that this problem has been tackled
> elsewhere, I could imagine some sort of get-entries-and-sth call?
>
Perhaps. As you say, though, once you've seen an STH for a tree of size N,
then you should expect to see a tree of that size. Waiting should resolve
the problem.

PR #51, opened 2015-06-12
Remove X509ChainEntry, clarify content of extra_data

mpalmer:
After `LogEntry` got nuked, I realised that `X509ChainEntry` was now also
unused as a structure.  So, I nuked that.  It was referenced in a couple of
places in the *text*, so I fixed those references up so they're now more
clear as to what they're *actually* talking about.  Since I needed something
to represent a certificate chain, I created `CertificateChain` as a
top-level structure.  I also cleaned up the structure definitions in a
couple of places:

* Once we've got `CertificateChain` defined, it's pretty clear that that's
  what the second element in PrecertChainEntryV2 is; and
* I didn't see any reason to have a case statement to choose between a
  `TBSCertificate` or a `TBSCertificate`, so I just made `signed_entry` in
  `SignedCertificateTimestamp` a straight-up member.

For anyone playing along at home, the definition of `extra_data` in
`/get-entries` is what tripped up alpha.ctlogs.org so badly, so I'm glad to
be able to clarify this for the future.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -382,16 +382,13 @@ following components:
 
     opaque ASN.1Cert&lt;1..2^24-1&gt;;
 
-    struct {
-        ASN.1Cert leaf_certificate;
-        ASN.1Cert certificate_chain&lt;0..2^24-1&gt;;
-    } X509ChainEntry;
+    ASN.1Cert CertificateChain&lt;0..2^24-1&gt;;
 
     opaque CMSPrecert&lt;1..2^24-1&gt;;
 
     struct {
         CMSPrecert pre_certificate;
-        ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
+        CertificateChain precertificate_chain;
     } PrecertChainEntryV2;
   </artwork>
 </figure>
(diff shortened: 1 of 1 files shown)


benlaurie:
In general, slightly confused by this change ... X509ChainEntry shows what the log needs to store - likewise a PrecertChainEntryV2. I think you are right that the structure is not actually used for anything, so there's certainly an argument for not showing it as a TLS structure, but if so, surely the same argument applies to PrecertChainEntryV2?

mpalmer:
I don't understand your confusion.  The `X509ChainEntry` data structure was not used anywhere; `PrecertChainEntryV2` *is* used.

robstradling:
"I didn't see any reason to have a case statement to choose between a TBSCertificate or a TBSCertificate"
Matt, the reason I'd left the case statement in the spec was that I figured there might be a future need to define new entry_type(s) that don't use TBSCertificate.

(Also, note that in TRAC ticket 80 we need to change these instances of TBSCertificate to something that also includes the Issuer Key Hash).

benlaurie:
@mpalmer I see what you mean. Looks to me like this is actually an error - the reason `PrecertChainEntryV2` is used is so the client can verify that the leaf was actually signed by the appropriate key. Now that we sign `TBSCertificate` for X.509 entries, too, we need to return the whole `X509ChainEntry` also to allow the same check.

robstradling:
@benlaurie Why would we need to "return the whole X509ChainEntry also to allow the same check" ?

Won't it be sufficient to just reintroduce the Issuer Key Hash to the SCT, as per TRAC ticket 80 ?

benlaurie:
@robstradling Because if you want to know who signed the TBSCertificate, you need to see the original certificate...

robstradling:
@benlaurie Ah I see.  The client you're talking about is a log monitor that needs get-entries to return the actual leaf certificate.  I agree with you that, in get-entries, "we need to return the whole X509ChainEntry also to allow the same check".
(Earlier I incorrectly thought that we were talking about a browser client that just needed to verify an SCT, having obtained the full cert chain from a TLS handshake.  Sorry for the pull-request-comment-spam ;-) )

robstradling:
@benlaurie I just created TRAC ticket 92 for the "we need to return the whole X509ChainEntry also to allow the same check" issue.

robstradling:
Closing this ticket.  I believe we've covered all of the issues mentioned.

PR #52, opened 2015-06-17
Deal with skew in get-entries.

benlaurie:
@pphaneuf @eranmes

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1243,7 +1243,7 @@ entry specified by <spanx style="verb">start</spanx>.
 	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
 	</t>
 	<t>
-	  In any case, the log server MUST return the latest STH it knows about in case of skew.
+	  In any case, the log server MUST return the latest STH it knows about when <spanx style="verb">start</spanx> is beyond that STH, and also when the last entry returned is the last entry corresponding to that STH. The log server MAY return the latest STH it knows at any other time.
 	</t>
       </section>
 
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1171,7 +1171,7 @@ an accepted root certificate.
 
       <section title="Retrieve Entries and STH from Log" anchor="fetch_entries">
         <t>
-          GET https://&lt;log server&gt;/ct/v1/get-entries
+          GET https://&lt;log server&gt;/ct/v2/get-entries
         </t>
         <t>
 	  <list style="hanging">
@@ -1243,7 +1243,7 @@ entry specified by <spanx style="verb">start</spanx>.
 	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
 	</t>
 	<t>
-	  In any case, the log server MUST return the latest STH it knows about when <spanx style="verb">start</spanx> is beyond that STH, and also when the last entry returned is the last entry corresponding to that STH. The log server MAY return the latest STH it knows at any other time.
+	  In any case, the log server MUST return the latest STH it knows about when <spanx style="verb">start</spanx> is beyond that STH or the last entry returned (which may be before <spanx style="verb">end</spanx>) is the last entry corresponding to that STH. The log server MAY return the latest STH it knows at any other time.
 	</t>
       </section>
 
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1243,7 +1243,7 @@ entry specified by <spanx style="verb">start</spanx>.
 	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
 	</t>
 	<t>
-	  In any case, the log server MUST return the latest STH it knows about when <spanx style="verb">start</spanx> is beyond that STH or the last entry returned (which may be before <spanx style="verb">end</spanx>) is the last entry corresponding to that STH. The log server MAY return the latest STH it knows at any other time.
+	  In any case, the log server MUST return the latest STH it knows about in case of skew.
 	</t>
       </section>
 
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1169,7 +1169,7 @@ an accepted root certificate.
 	</t>
       </section>
 
-      <section title="Retrieve Entries from Log" anchor="fetch_entries">
+      <section title="Retrieve Entries and STH from Log" anchor="fetch_entries">
         <t>
           GET https://&lt;log server&gt;/ct/v1/get-entries
         </t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1169,7 +1169,7 @@ an accepted root certificate.
 	</t>
       </section>
 
-      <section title="Retrieve Entries from Log" anchor="fetch_entries">
+      <section title="Retrieve Entries and STH from Log" anchor="fetch_entries">
         <t>
           GET https://&lt;log server&gt;/ct/v1/get-entries
```
eranmes: Update the version to v2?

eranmes:
Summary of a discussion with Ben yesterday:
- If I understand correctly, the main motivation behind this change is allowing a client to deal with skew by requiring a log front-end to return the latest STH it's aware of and all the entries up to it.
- Currently, if the end parameter is beyond the tree size a log front-end knows about it may return a partial response, but it may also refuse the request completely.
- This pull request will mandate this behaviour.
- By receiving a partial response, containing the latest STH known to the front-end and all entries up to it, the requesting client can verify the tree's consistency up to that point. Assuming the client is aware of a newer STH (since it requested  more entries that the front-end knows about) this means the client could drop the entries up to the received STH (after verifying tree consistency). The client would still have to hold onto the received STH to verify it is consistent with the newer STH it knows about.
- This is only useful if the response contains the entries up to the tree size in the returned STH. If the log decides to return less entries (for example, because the range requested was too big) then returning the STH has no use.

I think this mandates behaviour that is currently optional in favour of potential storange optimization on the client side (i.e. dropping the entries up to the returned STH). In practice the storage requirements for non-validated entries has not been a problem in the monitor  we've been running, so I'm not convinced this optimization justifies the added complexity to the protocol.

Maybe @robstradling has a different observation? Also @mpalmer because this PR is very similar to pr #47 he sent.

robstradling:
Eran, I'm also not convinced this optimization justifies the added complexity to the protocol.  But then, my monitor (https://crt.sh) doesn't drop entries.

If a log refuses to honour a request where (due to skew) end>tree_size, then the monitor could always try again later.

mpalmer:
My opinion is:

* Skew is a server-side problem.  If you publish an STH, but then don't serve the entries that are required to make up that STH, that's a Bad Thing.  To me, it feels *very* much like an MMD failure -- you're saying "yep, I've got that -- oh wait, no I don't!  Haha, fooled you!"
* Returning a 400 when you walk off the end of the log isn't great, although having a well-defined error to indicate specifically this occurence makes it a lot better than what 6962-compliant behaviour requires (which is 400 + no useful diagnostic information).

benlaurie:
@mpalmer The problem is one server can hand out the STH while another is not up-to-date. That second server could then be the one asked for the corresponding entries, which it does not have.

Whilst I think it is possible to engineer a server farm that doesn't do this, it:

a) is complicated, and

b) will hurt availability

@eranmes @robstradling whilst it is easiest to explain the need for this change in terms of a monitor that doesn't retain entries, it also seems to me that it is appropriate to always allow clients to verify data they hold (for example, to ensure there was no corruption). Without the corresponding STH this is not possible.

I do accept that there's some logic to only serving the STH when it applies to entries that have been served, so I will make that change.

rfc6962-bis.xml
```diff
@@ -1227,6 +1242,9 @@ entry specified by <spanx style="verb">start</spanx>.
 	<t>
 	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
 	</t>
+	<t>
+	  In any case, the log server MUST return the latest STH it knows about when <spanx style="verb">start</spanx> is beyond that STH, and also when the last entry returned is the last entry corresponding to that STH. The log server MAY return the latest STH it knows at any other time.
```
eranmes: Two nits:
(1) replace 'and also' with 'or' (if I understand correctly the latest STH MUST be returned when start > tree_size or last entry returned is tree_size).
(2) 'last entry returned' - so that can be smaller than the last entry requested, right?

benlaurie: On (2) yes, it can.

pphaneuf:
It's not really about optimising storage, either. I don't think a client should drop the entries past the STH that it received, but that STH allows it to validate the entries it *did* receive (integrate them in its local Merkle tree and verify the hash), while it keeps trying to get missing entries. If `get-entries` only returns a partial response, then you have no way of verifying that those entries are correct (because you're missing some entries relative to the STH you had, and so you can't calculate the hash).

Eran is correct that if `get-entries` returned fewer entries than requested because the range was too big, say, then the STH is not going to be useful immediately.

But my findings while poking around leads me to think that having more STHs (at different tree sizes) is convenient, as it gives clients "milestones" to validate progress while fetching entries. This is most obvious when mirroring a log like Pilot from nothing: you have an STH with 7.7M entries, but you won't have any way to detect corruption until you get the very last entry... If that hash doesn't match, the client has no idea *which* of the 7.7M has been corrupted (assuming no evil intentions here!)!

AlCutter:
On 23 Jun 2015 4:50 pm, "Pierre Phaneuf" <notifications@github.com> wrote:
>
> It's not really about optimising storage, either. I don't think a client
should drop the entries past the STH that it received, but that STH allows
it to validate the entries it did receive (integrate them in its local
Merkle tree and verify the hash), while it keeps trying to get missing
entries. If get-entries only returns a partial response, then you have no
way of verifying that those entries are correct (because you're missing
some entries relative to the STH you had, and so you can't calculate the
hash).
>
> Eran is correct that if get-entries returned fewer entries than requested
because the range was too big, say, then the STH is not going to be useful
immediately.
>
> But my findings while poking around leads me to think that having more
STHs (at different tree sizes) is convenient, as it gives clients
"milestones" to validate progress while fetching entries. This is most
obvious when mirroring a log like Pilot from nothing: you have an STH with
7.7M entries, but you won't have any way to detect corruption until you get
the very last entry... If that hash doesn't match, the client has no idea
which of the 7.7M has been corrupted (assuming no evil intentions here!)!

Yes it does, log2(N) carefully chosen inclusion proof requests against the
remote log will tell it which local leaf is corrupt, by comparing internal
tree nodes.

benlaurie:
Actually, on second thoughts, I don't think it is possible to engineer a server farm with no skew.

Anyway, @AlCutter: nice!

mpalmer:
A server which is out-of-date should not be answering requests.

I'm fairly certain that it *is* possible to engineer a server farm which is, at least, incapable of failing to serve a log entry covered by an STH which has been served.  One idea which comes immediately to mind is to generate the STH offline, ensure all members of the cluster have the entries covered by that STH (and, as a protection mechanism, cause the load balancers to drop any members which have failed to obtain the update), and only then cause the STH to be served.  I'm sure that other, more elegant, solutions could be developed.

pphaneuf:
Even when synchronizing the update of STH, you still get windows of inconsistency, these mechanisms only make them smaller. For example, the production Google log server produce an STH, and a timestamp at which to make it public, but clock skew can still give you inconsistent answers. Those are on a much smaller scale, though, so they are usually unnoticed, or easily fixed with a simple retry.

But in the case where infrastructure that we rely on for distribution of updated STH (and the timestamp at which they will become public) is compromised, you could get one of the nodes that does not know of the existence of this updated STH, and continues operating past the publication date of the new STH, not even knowing that it is out of date.

benlaurie:
As requested by @eranmes, reverted to a version that always returns the latest known STH, regardless of whether it is immediately useful.

Not entirely convinced this is an improvement, since the previous wording allowed a lazy log to do that anyway.

rfc6962-bis.xml
```diff
@@ -1227,6 +1242,9 @@ entry specified by <spanx style="verb">start</spanx>.
 	<t>
 	  Because of skew, it is possible the log server will not have any entries between <spanx style="verb">start</spanx> and <spanx style="verb">end</spanx>. In this case it MUST return an empty <spanx style="verb">entries</spanx> array.
 	</t>
+	<t>
+	  In any case, the log server MUST return the latest STH it knows about in case of skew.
```
eranmes: Nit: Add that a log server MAY return the latest STH in all other cases?

benlaurie: ENITNOTUNDERSTOOD - the STH MUST be returned in ALL cases.

eranmes:
As discussed offline, LGTM once you eliminate the "in case of skew.".

benlaurie:
Merged at b94e659.

PR #53, opened 2015-06-25
Also return SCTs.

benlaurie:
@eranmes @pphaneuf http://trac.tools.ietf.org/wg/trans/trac/ticket/82.

Note that this uses the binary format for the SCT, as we've decided will be used elsewhere.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1196,6 +1196,9 @@ an accepted root certificate.
                     <t hangText="extra_data:">
                       The base64-encoded unsigned data pertaining to the log entry. In the case of an X509ChainEntry, this is the <spanx style="verb">certificate_chain</spanx>. In the case of a PrecertChainEntryV2, this is the whole <spanx style="verb">PrecertChainEntryV2</spanx>.
                     </t>
+		    <t hangText="sct:">
+		      A base64-encoded <spanx style="verb">SignedCertificateTimestamp</spanx> for this entry. Note that more than one SCT may have been returned for the same entry - only one of those is returned in this field. It may not be possible to retrieve others.
+		    </t>
 		  </list>
 		</t>
               </list>


rfc6962-bis.xml
```diff
@@ -1196,6 +1196,9 @@ an accepted root certificate.
                     <t hangText="extra_data:">
                       The base64-encoded unsigned data pertaining to the log entry. In the case of an X509ChainEntry, this is the <spanx style="verb">certificate_chain</spanx>. In the case of a PrecertChainEntryV2, this is the whole <spanx style="verb">PrecertChainEntryV2</spanx>.
                     </t>
+		    <t hangText="sct:">
+		      A base64-encoded <spanx style="verb">SignedCertificateTimestamp</spanx> for this entry. Note that more than one SCT may have been returned for the same entry - only one of those is returned in this field. It may not be possible to retrieve others.
```
eranmes: I assume the encoding is the TLS encoding? If so, may be worth pointing out this is base64-encoded.

eranmes:
LGTM, see the one comment about encoding.

rfc6962-bis.xml
benlaurie: EALSONOTUNDERSTOOD. The second word is "base64-encoded", plus the SignedCertificateTimstamp structure is a TLS encoded structure, by definition.

benlaurie:
Merged at 9141ac7.

rfc6962-bis.xml
pphaneuf: How is it even possible to have more than one SCT for a given log index?!? For a hash, yes, a given certificate might have been submitted more than once, but `get-entries` is index-based, should map to exactly one SCT, no?

eranmes: I believe you're correct as the timestamp is included in the entry.

pphaneuf:
Shouldn't `/ct/v1/get-entries` be `/ct/v2/get-entries`?

PR #57, opened 2015-07-03
Add algorithms for verifying inclusion proofs, consistency checks

aeijdenberg:
Some other misc spacing changes to remove warnings from xml2rfc tool.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -378,21 +378,21 @@ following components:
         </t>
 <figure>
   <artwork>
-   enum { x509_entry(0), precert_entry_V2(3), (65535) } LogEntryType;
+    enum { x509_entry(0), precert_entry_V2(3), (65535) } LogEntryType;
 
-   opaque ASN.1Cert&lt;1..2^24-1&gt;;
+    opaque ASN.1Cert&lt;1..2^24-1&gt;;
 
-   struct {
-       ASN.1Cert leaf_certificate;
-       ASN.1Cert certificate_chain&lt;0..2^24-1&gt;;
-   } X509ChainEntry;
+    struct {
+        ASN.1Cert leaf_certificate;
+        ASN.1Cert certificate_chain&lt;0..2^24-1&gt;;
+    } X509ChainEntry;
 
-   opaque CMSPrecert&lt;1..2^24-1&gt;;
+    opaque CMSPrecert&lt;1..2^24-1&gt;;
 
-   struct {
-       CMSPrecert pre_certificate;
-       ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
-   } PrecertChainEntryV2;
+    struct {
+        CMSPrecert pre_certificate;
+        ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
+    } PrecertChainEntryV2;
   </artwork>
 </figure>
         <t>
@@ -791,16 +791,16 @@ signatures.
 
 	<figure>
 	  <artwork>
-            enum { v1(0), (255) } TreeHeadVersion;
+	    enum { v1(0), (255) } TreeHeadVersion;
 
-            digitally-signed struct {
+	    digitally-signed struct {
               TreeHeadVersion version;
               SignatureType signature_type = tree_hash;
               uint64 timestamp;
               uint64 tree_size;
               opaque sha256_root_hash[32];
-            } TreeHeadSignature;
-          </artwork>
+	    } TreeHeadSignature;
+	  </artwork>
 	</figure>
 
         <t>
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1444,7 +1444,7 @@ the STH changes.
                     <vspace blankLines="0" />
                     Set <spanx style="verb">sr</spanx> to <spanx style="verb">HASH(0x01 || c || sr)</spanx>
                   </t>
-                  <t>If <spanx style="verb">LSB(fn)</spanx> is not set, then right-shift both <spanx style="verb">fn</spanx> and <spanx style="verb">fn</spanx> equally until <spanx style="verb">LSB(fn)</spanx> is set.</t>
+                  <t>If <spanx style="verb">LSB(fn)</spanx> is not set, then right-shift both <spanx style="verb">fn</spanx> and <spanx style="verb">sn</spanx> equally until <spanx style="verb">LSB(fn)</spanx> is set.</t>
                 </list>
                 Otherwise:
                 <list style="empty">
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -123,7 +123,7 @@ Section 4 of <xref target="RFC5246"/>.
       </section>
     </section>
     <section title="Cryptographic Components">
-      <section title="Merkle Hash Trees" anchor="merkle_hash_trees">
+      <section title="Merkle Hash Trees">
         <t>
           Logs use a binary Merkle Hash Tree for efficient auditing. The hashing algorithm used by each log is expected to be specified as part of the metadata relating to that log. We have established a registry of acceptable algorithms, see <xref target="hash_algorithms"/>. The hashing algorithm in use is referred to as HASH throughout this document. The input to the Merkle Tree Hash is a list
 of data entries; these entries will be hashed to form the leaves of the Merkle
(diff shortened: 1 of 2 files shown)


rfc6962-bis.xml
```diff
@@ -378,21 +378,21 @@ following components:
...
+   opaque CMSPrecert&lt;1..2^24-1&gt;;
 
-    struct {
-        CMSPrecert pre_certificate;
-        ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
-    } PrecertChainEntryV2;
+   struct {
+       CMSPrecert pre_certificate;
+       ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
+   } PrecertChainEntryV2;
```
benlaurie: Why the indentation change?

rfc6962-bis.xml
```diff
@@ -1015,10 +1015,10 @@ an accepted root certificate.
 		  The Merkle Tree Hash of the tree, in base64.
 		</t>
 		<t hangText="tree_head_signature:">
-		  A TreeHeadSignature for the above data.
+		  A TreeHeadSignature for <spanx style="verb">tree_size</spanx>, <spanx style="verb">timestamp</spanx> and <spanx style="verb">sha256_root_hash</spanx>.
```
benlaurie: This appears to be #56

rfc6962-bis.xml
```diff
@@ -1396,6 +1402,94 @@ the STH changes.
...
+                For each subsequent value <spanx style="verb">c</spanx> in the <spanx style="verb">consistency</spanx> array:
+                <vspace blankLines="1" />
+                If <spanx style="verb">LSB(fn)</spanx> is set, or if <spanx style="verb">fn</spanx> is equal to <spanx style="verb">sn</spanx>, then:
+                <list style="numbers">
+                  <t>
+                    Set <spanx style="verb">fr</spanx> to <spanx style="verb">HASH(0x01 || c || fr)</spanx>
+                    <vspace blankLines="0" />
+                    Set <spanx style="verb">sr</spanx> to <spanx style="verb">HASH(0x01 || c || sr)</spanx>
+                  </t>
+                  <t>If <spanx style="verb">LSB(fn)</spanx> is not set, then right-shift both <spanx style="verb">fn</spanx> and <spanx style="verb">fn</spanx> equally until <spanx style="verb">LSB(fn)</spanx> is set.</t>
```
benlaurie: both fn and fn?

benlaurie: !LSB(fn) => fn == sn. So the effect of this appears to be to apply the remaining values in the same way to both sn and fn, which means that fn == sn at the end. So, seems like you could terminate here.

rfc6962-bis.xml
```diff
@@ -1396,6 +1402,94 @@ the STH changes.
...
+                </list>
+                Otherwise:
+                <list style="empty">
+                  <t>Set <spanx style="verb">sr</spanx> to <spanx style="verb">HASH(0x01 || sr || c)</spanx></t>
+                </list>
+                <vspace blankLines="1" />
+                Finally, right-shift both <spanx style="verb">fn</spanx> and <spanx style="verb">sn</spanx> one time.
+              </t>
+              <t>After completing iterating through the <spanx style="verb">consistency</spanx> array as described above, verify that the <spanx style="verb">fr</spanx> calculated is equal to the <spanx style="verb">first_hash</spanx> supplied and that the <spanx style="verb">sr</spanx> calculated is equal to the <spanx style="verb">second_hash</spanx> supplied.</t>
```
benlaurie: Presumably you should also check that:

a) fn == sn == 0.

b) all entries in |consistency| have been used.

rfc6962-bis.xml
aeijdenberg: Unless my system is unique, xml2rfc gives a warning saying that it is outdenting this artwork by one space so that it fits.
I was sick of seeing the warning so I manually outdented each.  The warning stops and as far as I can tell the output is identical.

rfc6962-bis.xml
aeijdenberg: Yes, I haven't really done this before.  I guess I should have made a separate branch for each from master? (Instead I branched after created the first PR).

aeijdenberg: Note the other change has been accepted now.

rfc6962-bis.xml
aeijdenberg: Corrected 2nd fn to sn.  Let me think about the other statement.

rfc6962-bis.xml
aeijdenberg: b) is guaranteed because we're iterating through the consistency array (this differs from our open-source code which does it differently).

a) "fn" and "sn" are means to an end. Does it matter if they are zero or not so long as the hashes calculated both match?  Since we iterate through all elements of the consistency array and always apply to at least "sr" I would expect that comparing "sr" against the supplied hash to be sufficient to verify correct consistency date is supplied.

rfc6962-bis.xml
aeijdenberg: The loop is over the values in the consistency array - as such I can't see how we could terminate early since all values are needed to produce the second hash.

rfc6962-bis.xml
aeijdenberg: Moved to separate PR.

rfc6962-bis.xml
```diff
@@ -1398,6 +1402,86 @@ the STH changes.
...
+                For each subsequent value <spanx style="verb">c</spanx> in the <spanx style="verb">consistency</spanx> array:
+                <vspace blankLines="1" />
+                If <spanx style="verb">LSB(fn)</spanx> is set, or if <spanx style="verb">fn</spanx> is equal to <spanx style="verb">sn</spanx>, then:
+                <list style="numbers">
+                  <t>
+                    Set <spanx style="verb">fr</spanx> to <spanx style="verb">HASH(0x01 || c || fr)</spanx>
+                    <vspace blankLines="0" />
+                    Set <spanx style="verb">sr</spanx> to <spanx style="verb">HASH(0x01 || c || sr)</spanx>
+                  </t>
+                  <t>If <spanx style="verb">LSB(fn)</spanx> is not set, then right-shift both <spanx style="verb">fn</spanx> and <spanx style="verb">sn</spanx> equally until <spanx style="verb">LSB(fn)</spanx> is set.</t>
```
benlaurie: If the log is evil/broken, then this could be an infinite loop.

rfc6962-bis.xml
benlaurie: If fn and sn are not zero at the end, then something weird has happened, and the hashes should presumably not match, so it may indeed be redundant, but might be worth mentioning as a sanity check.

rfc6962-bis.xml
aeijdenberg: Gotcha, if the consistency proof returned has too many entries, both fn and sn will reduce to zero and get stuck here.

rfc6962-bis.xml
```diff
@@ -123,7 +123,7 @@ Section 4 of <xref target="RFC5246"/>.
       </section>
     </section>
     <section title="Cryptographic Components">
-      <section title="Merkle Hash Trees">
+      <section title="Merkle Hash Trees" anchor="merkle_hash_trees">
```
benlaurie: This should not be here.

rfc6962-bis.xml
```diff
@@ -781,7 +781,7 @@ signatures.
           <spanx style="verb">extensions</spanx> are <spanx style="verb">extensions</spanx> of the corresponding SCT.
         </t>
         <t>
-          The leaves of the Merkle Tree are the leaf hashes of the corresponding <spanx style="verb">MerkleTreeLeaf</spanx> structures.
+          The leaves of the Merkle Tree are the leaf hashes of the corresponding <spanx style="verb">MerkleTreeLeaf</spanx> structures. Note that leaf hashes are calculated as <spanx style="verb">HASH(0x00 || MerkleTreeLeaf)</spanx>. See <xref target="merkle_hash_trees"/> for further details.
```
benlaurie: Nor this.

rfc6962-bis.xml
```diff
@@ -1236,6 +1239,7 @@ entry specified by <spanx style="verb">start</spanx>.
 	<t>
 	  In any case, the log server MUST return the latest STH it knows about.
 	</t>
+        <t>See <xref target="verify_hash" /> for an outline of how to use a complete list of <spanx style="verb">leaf_input</spanx> entries to verify the <spanx style="verb">sha256_root_hash</spanx>.</t>
```
benlaurie: sha256_root_hash is now root_hash.

benlaurie:
LGTM for now. Better tests needed.

PR #58, opened 2015-07-03
Ticket 80: Re-introduce the issuer key hash into the SCT.

robstradling:


rfc6962-bis.xml
```diff
@@ -557,6 +562,9 @@ SEQUENCE {
           <spanx style="verb">key_id</spanx> is the SHA-256 hash of the log's public key, calculated over the DER encoding of the key represented as SubjectPublicKeyInfo.
         </t>
         <t>
+          <spanx style="verb">issuer_key_hash</spanx> is the SHA-256 hash of the certificate issuer's public key, calculated over the DER encoding of the key represented as SubjectPublicKeyInfo. This is needed to bind the issuer to the final certificate.
```
benlaurie: More properly, I think, is to "bind the issuer, as defined by OCSP, to the final certificate".

Too lazy to check what CRLs need right now.

robstradling: FWIW, I simply restored the issuer_key_hash definition from RFC6962.  :-)

In an imaginary world where all certs that share the same TBSCertificate are guaranteed to be reported as either all revoked or all unrevoked, this pull request would not be necessary.  Since that's not the world we live in, it's necessary for CT to detect the issuance of every certificate, not just every TBSCertificate.  Therefore, each SCT MUST be valid for precisely 1 certificate, because only 1 certificate needs to be logged in order for the SCT to be produced and made public.  Binding only to the Issuer Name is insufficient, since the same Issuer Name could be used by multiple CAs in multiple certificate chains.

Since revocation is out of scope for CT, I don't think it makes sense to insert "as defined by OCSP".  How about we change that sentence to:
"This is needed to bind the issuer to the final certificate, making it impossible for the SCT to be valid for any other certificate."
?

benlaurie: I just wanted a reminder why we need this.

Although CT does not deal with revocation, clearly there is little point detecting mis-issuance if there's nothing to be done about it, which is the relevance of OCSP.

I guess your suggested text sort of does that. So, with that mod, I'm OK with this update.

robstradling:
Merged at f2650ef2b776a05eb280614b967efe2fcc4cf44e

PR #64, opened 2015-07-06
Use LogEntryType 2.

robstradling:
RFC6962 only uses 0 and 1.
We almost used 2 for an x509_entry_V2, but later concluded that we didn't need it (hence the current use of 3 rather than 2 for precert_entry_V2).
Let's use 2 rather than skip over it unnecessarily.


robstradling:
Merged at 0bc8a18cde62cdbeb23ffe1a4dae98b291d8c249

PR #70, opened 2015-07-17
Ticket 70: Define top-level extension syntax (text by Steve Kent)

robstradling:


rfc6962-bis.xml
```diff
@@ -555,7 +555,16 @@ SEQUENCE {
           TBSCertificate tbs_certificate;
       } CertInfo;
 
-      opaque CtExtensions&lt;0..2^16-1&gt;;
+    uint16 CtExtensionId;
```
robstradling: TLS 1.2 hello extension IDs have an associated IANA registry but they're defined as an enum rather than a uint16.  See RFC5246 Section 7.4.1.4.

We should do the same here.

rfc6962-bis.xml
```diff
@@ -572,6 +581,15 @@ TBSCertificate from it, redacting the domain name labels indicated by the
 redacted labels extension, and deleting the SCT list extension and redacted
 labels extension.
         </t>
+        <t>
+          <spanx style="verb">id</spanx> identifies a single extension from the IANA registry in <xref target="extension_ids"/>.
```
robstradling: For clarity, how about: s/id/CtExtension.id/

rfc6962-bis.xml
```diff
@@ -572,6 +581,15 @@ TBSCertificate from it, redacting the domain name labels indicated by the
...
+        </t>
+        <t>
+          The interpretation of the <spanx style="verb">value</spanx> field is determined solely by the value of the <spanx style="verb">id</spanx> field, and each document that registers a new extension ID must describe how to interpret the corresponding extension value.
```
robstradling: Ditto.  Change value to CtExtension.value and id to CtExtension.id.

rfc6962-bis.xml
```diff
@@ -610,8 +628,7 @@ protocol to which the SCT conforms. This version is v2. Note that <xref target="
           <spanx style="verb">signed_entry</spanx> includes the TBSCertificate from either the <spanx style="verb">leaf_certificate</spanx> (in the case of an X509ChainEntry) or the <spanx style="verb">pre_certificate</spanx> (in the case of a PrecertChainEntryV2).
         </t>
         <t>
-          <spanx style="verb">extensions</spanx> are future extensions to
-SignedCertificateTimestamp v2. Currently, no extensions are specified.
+          <spanx style="verb">extensions</spanx> are future extensions to SignedCertificateTimestamp v2. Currently, no extensions are specified. If an implementation sees an extension that it does not understand, it SHOULD ignore that extension. Furthermore, an implementation MAY choose to ignore any extension(s) that it does understand.
```
robstradling: Should we also explain that if there's ever a need to define an SCT "extension" that clients MUST understand and process, the way to accommodate this will be to bump the SCT version and change the SCT structure?

robstradling:
@benlaurie @eranmes

rfc6962-bis.xml
benlaurie: We still have to decide the width, though.

robstradling: Sure.  I propose:

enum {
    reserved(65535)
} CtExtensionType;

rfc6962-bis.xml
benlaurie: Inconsistent with everything else in the document, so I say no.

rfc6962-bis.xml
benlaurie: Again, no.

rfc6962-bis.xml
benlaurie: Or should we add the notion of critical extensions?

rfc6962-bis.xml
benlaurie: +1

rfc6962-bis.xml
robstradling: I'm a little bit concerned about possible confusion with "LogID id;", which is defined in the same section.  But I'll think about addressing this in a separate PR.

robstradling: Actually, inspired by reviewing RFC5246 Section 7.4.1.4 again just now, I think we should do this...

    enum {
        reserved(65535)
    } CtExtensionType;

    struct {
        CtExtensionType extension_type;
        opaque extension_data<0..2^16-1>;
    } CtExtension;

    CtExtension extensions<0..2^16-1>;

OK?

benlaurie: +1

robstradling:
Merged at 91a7ce0bc4e1b51e9e22d015cd6fefc0fcebd202 and 7525741318ee7cd4f79d126a708522482b81b1f5

rfc6962-bis.xml
29simonka29: https://github.com/google/certificate-transparency-rfcs/pull/70#discussion_r34991818
Simona Pinkova

PR #75, opened 2015-07-23
Describe a log is extensible, not necessarily ever-growing.

eranmes:
A log's size may remain constant (in case it is frozen) or increase (in case
of normal operation). Hopefully that better describes it.


rfc6962-bis.xml
```diff
@@ -311,7 +311,7 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
         Anyone can submit certificates to certificate logs for public auditing;
 however, since certificates will not be accepted by TLS clients unless logged,
 it is expected that certificate owners or their CAs will usually submit them. A
-log is a single, ever-growing, append-only Merkle Tree of such certificates.
+log is a single, extensible, append-only Merkle Tree of such certificates.
```
benlaurie: Doesn't append-only imply extensible? Suggest just drop the word altogether.

PR #77, opened 2015-07-23
Add the maximum chain length to the log metadata.

eranmes:
Since logs SHOULD limit the length of chains accepted, that value should be
known to clients in advance.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -394,7 +394,7 @@ certificate. Logs MUST also reject precertificates that are not valid DER encode
     } PrecertChainEntryV2;</artwork>
         </figure>
         <t>
-          Logs SHOULD limit the length of chain they will accept.
+          Logs SHOULD limit the length of chain they will accept. The maximum chain length is specified in the log's metadata.
         </t>
         <t>
           <spanx style="verb">entry_type</spanx> is the type of this entry. Future revisions of this protocol may add new LogEntryType values. <xref target="client_messages"/> explains how clients should handle unknown entry types.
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -394,7 +394,7 @@ certificate. Logs MUST also reject precertificates that are not valid DER encode
     } PrecertChainEntryV2;</artwork>
         </figure>
         <t>
-          Logs SHOULD limit the length of chain they will accept.
+          Logs SHOULD limit the length of chain they will accept (which is a part of the log's metadata).
```
robstradling: The grammar seems a bit clunky there.  How about:
"Logs SHOULD limit the length of chain they will accept.  The maximum chain length is specified in the log's metadata."

rfc6962-bis.xml
```diff
@@ -1321,7 +1321,10 @@ but it is expected there will be a variety.
 	  </t>
 	  <t hangText="Version">
 	    The version of the protocol supported by the log (currently 1 or 2).
-	  </t>
+    </t>
+    <t hangText="Maximum Chain Length">
+      The longest chain submission this log is willing to accept.
```
robstradling: s/this/the/  for consistency with the descriptions of the other metadata items.

robstradling: "Logs SHOULD limit..." is not a MUST, which implies that there needs to be a way to _not_ specify a limit.  Is this done by omitting, or by specifying some magic value for, the "Maximum Chain Length" metadata item?

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: By omission - I'll amend the text to reflect.

eranmes:
Pushed a new version, PTAL.

rfc6962-bis.xml
```diff
@@ -1321,7 +1321,10 @@ but it is expected there will be a variety.
 	  </t>
 	  <t hangText="Version">
 	    The version of the protocol supported by the log (currently 1 or 2).
-	  </t>
+    </t>
+    <t hangText="Maximum Chain Length">
+      The longest chain submission the log is willing to accept, if the log chose to limit it.
+    </t>
```
benlaurie: Indent is completely wrong.

Do we need to say "in decimal"?

eranmes: Fixed indentation.
I don't think we need to say it's in decimal - the metadata format is not strictly specified since how it is distributed is client-specific.

PR #78, opened 2015-07-24
Bashing more client behaviour statements.

eranmes:
The RFC should specify how to determine whether a certificate is CT-compliant
or not, by specifying the requirement for SCTs to be present alongside it
during the TLS handshake and how to check their validity.
Once the compliance status of the certificate has been established, it is
up to the client to decide how it behaves.

This commit attempts to reflect that.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -96,8 +96,7 @@ can also contribute their own certificate chains, as can third parties. In order
 rendered useless by submitting large numbers of spurious certificates, it is required that each chain is rooted in a
 CA certificate accepted by the log. When a chain is submitted to a log, a signed timestamp is
 returned, which can later be used to provide evidence to TLS clients that the chain
-has been submitted. TLS clients can thus require that all certificates they accept as valid
-have been logged.
+has been submitted. TLS clients can thus require that all certificates they accept as valid are accompanied by signed timestamps.
       </t>
       <t>
         Those who are concerned about misissue can monitor the logs, asking
@@ -653,10 +652,7 @@ updated on the fly.
             </t>
             <t>
               An X509v3 certificate extension
-(see <xref target="cert_sctlist_extension"/>). This mechanism allows the use of
-unmodified TLS servers, but, because the included SCTs cannot be changed without
-re-issuing the certificate, increases the risk that the certificate will be
-refused if any of the SCTs become invalid.
+(see <xref target="cert_sctlist_extension"/>). This mechanism allows the use of unmodified TLS servers, but the SCTs cannot be updated on the fly. Since the logs that signed the SCTs won't necessarily be accepted by TLS clients for the full lifetime of the certificate, there is a risk that TLS clients will subsequently consider the certificate to be non-compliant and in need of re-issuance.
             </t>
           </list>
         </t>
@@ -1285,7 +1281,7 @@ entry specified by <spanx style="verb">start</spanx>.
     <section title="Clients">
       <t>
         There are various different functions clients of logs might perform. We
-describe here some typical clients and how they could function. Any
+describe here some typical clients and how they should function. Any
 inconsistency may be used as evidence that a log has not behaved correctly, and
 the signatures on the data structures prevent the log from denying that
 misbehavior.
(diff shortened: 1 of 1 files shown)


eranmes:
soliciting feedback from @robstradling as well.

rfc6962-bis.xml
```diff
@@ -656,7 +655,7 @@ updated on the fly.
 (see <xref target="cert_sctlist_extension"/>). This mechanism allows the use of
 unmodified TLS servers, but, because the included SCTs cannot be changed without
 re-issuing the certificate, increases the risk that the certificate will be
-refused if any of the SCTs become invalid.
+considered non-compliant if some of the SCTs become invalid (by the client not recognizing the logs they were issued from anymore).
```
robstradling: I've had a go at re-wording this paragraph.  How's this?
"This mechanism allows the use of unmodified TLS servers, but the SCTs cannot be updated on the fly.  Since the logs that signed the SCTs won't necessarily be accepted by TLS clients for the full lifetime of the certificate, there is a risk that TLS clients will subsequently consider the certificate to be non-compliant with this document and in need of re-issuance."

rfc6962-bis.xml
```diff
@@ -1554,7 +1553,7 @@ the STH changes.
       </t>
       <section title="Misissued Certificates">
         <t>
-          Misissued certificates that have not been publicly logged, and thus do not have a valid SCT, will be rejected by TLS clients. Misissued certificates that do have an SCT from a log will appear in that public log within the Maximum Merge Delay, assuming the log is operating correctly. Thus, the maximum period of time during which a misissued certificate can be used without being available for audit is the MMD.
+          Misissued certificates that have not been publicly logged, and thus do not have a valid SCT, are not considered compliant with this document (So TLS clients may decide, for example, to reject them). Misissued certificates that do have an SCT from a log will appear in that public log within the Maximum Merge Delay, assuming the log is operating correctly. Thus, the maximum period of time during which a misissued certificate can be used without being available for audit is the MMD.
```
robstradling: Nit: s/(So/(so/

rfc6962-bis.xml
benlaurie: In general I prefer "non-compliant" to "non-compliant with this document". If we really need to, we could add a note at the top somewhere saying that by "non-compliant" we mean "non-compliant with this document", though I'd think that was obvious.

That said, I otherwise prefer Rob's version.

rfc6962-bis.xml
```diff
@@ -1344,7 +1343,7 @@ but it is expected there will be a variety.
 addition to normal validation of the certificate and its chain, TLS clients
 SHOULD validate the SCT by computing the signature input from the SCT data as
 well as the certificate and verifying the signature, using the corresponding
-log's public key.
+log's public key. By validating SCTs, TLS clients can thus determine whether certificates are compliant with this document. However, specifying the TLS clients' behaviour once compliance or non-compliance has been determined (for example, whether a certificate should be rejected due to the lack of valid SCTs) is outside the scope of this document.
```
benlaurie: Same comment about compliant.

rfc6962-bis.xml
benlaurie: And again.

rfc6962-bis.xml
robstradling: Note that my version also removed the notion of SCTs "becoming invalid".  The signature on an SCT doesn't cease to be valid just because the log that signed the SCT is no longer accepted by the client.

eranmes: Done - incorporated Rob's suggested text, dropping "with this document".
@benlaurie , any suggestion on where (in the document) to define compliance, if at all?

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: Done - addressed both comments.

rfc6962-bis.xml
robstradling: Doesn't section 1.1 explain "compliance" clearly enough already?

1.1.  Requirements Language

   The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
   "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
   document are to be interpreted as described in RFC 2119 [RFC2119].

eranmes: I think so; I've added "with this document" due to complaints in another trac ticket, but if we're all in agreement it's superfluous then we can leave it this way and re-open the discussion if anybody brings up the issue of definition of compliance with the new wording.

PR #80, opened 2015-09-15
Include LogID in the STH structure (TreeHeadSignature)

robstradling:


robstradling:
Does get-sth need a version bump from /v1/get-sth to /v2/get-sth as a result of this change?

pphaneuf:
@eranmes was telling me that *all* the methods were going to be bumped to v2 in their path, anyway?

robstradling:
@pphaneuf I just chatted to Ben: he agrees that we should bump get-sth to v2 now.  get-roots is the only method still on v1 at the moment.

PR #81, opened 2015-09-15
Ticket 5: Add extensions to STH.

robstradling:


rfc6962-bis.xml
```diff
@@ -835,6 +858,9 @@ signatures.
         <t>
           <spanx style="verb">root_hash</spanx> is the root of the Merkle Hash Tree.
         </t>
+        <t>
+          <spanx style="verb">extensions</spanx> are future extensions to STH v2. Currently, no extensions are specified. If an implementation sees an extension that it does not understand, it SHOULD ignore that extension. Furthermore, an implementation MAY choose to ignore any extension(s) that it does understand.
```
benlaurie: I presume this matches the SCT text, but I wonder ... should we leave it to the extension whether it can be ignored if understood? Though I guess in practice its up to the implementor to decide what is "understood".

robstradling: Yes, this text matches the SCT text.

Maybe we should just remove the "Furthermore...that it does understand" sentence entirely?

benlaurie: I'm easy either way, TBH.

robstradling: OK.  I'll keep that sentence.  Thanks.

robstradling:
Merged at 81c9479b155916db0cf91e86a76d40192ea2750a

PR #83, opened 2015-09-21
Ticket 107: Struct-ify STH

robstradling:


rfc6962-bis.xml
```diff
@@ -824,45 +825,52 @@ signatures.
...
+            opaque root_hash[HASH_SIZE];
+            SthExtensions extensions;
+        };
+    } SignedTreeHead;</artwork>
+         </figure>
+          <t>
+            <spanx style="verb">version</spanx> is the version of the SignedTreeHead structure. This version is v2. Note that <xref target="RFC6962">TreeHeadSignature v1</xref> only included the inner <spanx style="verb">digitally-signed struct</spanx> and did not include the <spanx style="verb">id</spanx> or <spanx style="verb">extensions</spanx> fields.
+          </t>
+          <t>
+            <spanx style="verb">timestamp</spanx> is the current time. The timestamp MUST be at least as recent as the most recent SCT timestamp in the tree. Each subsequent timestamp MUST be more recent than the timestamp of the previous update.
```
benlaurie: For SCTs we define what we mean by a timestamp, we should probably do the same here.

robstradling: FWIW, this "timestamp" definition is currently unchanged since 6962.  But yes, let's take this opportunity to make it more precise.  I'll copy the "...is the current NTP Time..." bits from the definition of the SCT timestamp.

robstradling:
Merged at d29431143d7bc47d1f9030ba8f6883353b698afa

rfc6962-bis.xml
```diff
@@ -801,15 +801,16 @@ signatures.
...
         <t>
           Periodically the log SHOULD sign the corresponding tree hash and tree information (see the corresponding Signed Tree Head client message in <xref target="fetch_sth"/>).
         </t>
-        <figure>
-          <preamble>
-            The signature for that data is structured as follows:
-          </preamble>
-          <artwork>
+        <t>
+          Each log MUST produce on demand a Signed Tree Head that is no older than the Maximum Merge Delay. However, Signed Tree Heads could be used to mark individual clients (by producing a new one for each query), so logs MUST NOT produce them more frequently than is declared in their metadata. In general, there is no need to produce a new Signed Tree Head unless there are new entries in the log, however, in the unlikely event that it receives no new submissions during an MMD period, the log SHALL sign the same Merkle Tree Hash with a fresh timestamp.
```
eranmes: Nit: The text says the log SHALL sign the same MTH with a fresh timestamp but does not say when.

robstradling: That's true, but that text wasn't introduced by this change.  Let's tackle that in a new PR.

PR #86, opened 2015-10-19
Reorganize requirements into appropriate sections

robstradling:
The Log Format and Operation section should describe just that.  This PR moves requirements for other parties (Submitters, CAs, TLS servers and TLS clients) into more appropriate sections.


eranmes:
LGTM.

PR #87, opened 2015-10-19
Ticket 105: Shrink LogID.

robstradling:
Putting the "MUST NOT use the same keypair" requirement in the Metadata section doesn't seem ideal, but it's the best place I can find to hang it for now.


rfc6962-bis.xml
```diff
@@ -541,9 +541,7 @@ ranges.
         v2(1), (255)
     } Version;
 
-    struct {
-        opaque key_id[HASH_SIZE];
-    } LogID;
+    opaque LogID&lt;1..2^8-1&gt;;
```
eranmes: suggest increasing to two bytes as there are already ~10 logs, we'd like to be able to run test logs as well.

rfc6962-bis.xml
```diff
@@ -614,6 +607,9 @@ labels extension.
 protocol to which the SCT conforms. This version is v2. Note that <xref target="RFC6962">SignedCertificateTimestamp v1</xref> had a different definition of <spanx style="verb">signed_entry</spanx>.
         </t>
         <t>
+          <spanx style="verb">log_id</spanx> is the DER encoding of an OID, excluding the ASN.1 tag and length bytes, that the log operator has allocated for the purpose of uniquely identifying this log. This OID is specified in the log's metadata.
```
eranmes: Dropping the ASN.1 tag and length bytes implies the length byte would be fixed, correct?

rfc6962-bis.xml
robstradling: That <1..2^8-1> means 1 byte is used to hold the variable length number of bytes of the DER-encoded OID.  I've never seen a DER-encoded OID use up anywhere near 255 bytes!

rfc6962-bis.xml
robstradling: No.  As explained above, the <1..2^8-1> holds a variable length.  The ASN.1 length can be easily derived from the <1..2^8-1> length byte.

rfc6962-bis.xml
eranmes: Oh, I see - it's an opaque field of 1 - 255 octets.
Not sure how that fits with the DER encoding? i.e. should it / could it be converted to an ASN.1 OID and if so, how?

robstradling: The OBJECT IDENTIFIER octet and the ASN.1 length octet(s) are stripped from the DER-encoded OID, leaving just the ASN.1 data octets.  Those ASN.1 data octets are what is stored in the opaque field of 1 - 255 octets.
If needed, the DER-encoded OID can be easily reconstructed from just those data octets: the ASN.1 tag octet is always 0x06; and if the length of the opaque field is <=127 octets (which is true for 100% of the OIDs I've even seen!), the ASN.1 length octet will be identical to the opaque field length.  (In other words, you can convert between the two forms by prepending a 0x06 octet or removing the leading 0x06 octet).

robstradling:
Merged at 22746c1a6197c665c7e8ff5d8008c5ab9c7c3537

PR #88, opened 2015-10-20
Change the LogID length to fit into one byte.

eranmes:
That yields a valid DER-encoded ASN.1 OID by prepending the tag 0x06 to
the LogID bytes.


eranmes:
Replaced my suggested text with Rob's - PTAL.

PR #89, opened 2015-10-31
Tickets 10 and 104: Enable inclusion proofs to appear in extensions, etc, etc

robstradling:


rfc6962-bis.xml
```diff
@@ -419,7 +419,7 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
         A log is a single, append-only Merkle Tree of submitted certificate and precertificate entries.
       </t>
       <t>
-        When it receives a valid submission, the log MUST return an SCT that corresponds to the submitted certificate or precertificate. If the log has previously seen this valid submission, it MAY return the same SCT as it returned before (note that if a certificate was previously logged as a precertificate, then the precertificate's SCT would not be appropriate, instead a fresh SCT of type x509_entry should be generated).
+        When it receives a valid submission, the log MUST return an SCT that corresponds to the submitted certificate or precertificate. If the log has previously seen this valid submission, it MAY return the same SCT as it returned before. (Note that if a certificate was previously logged as a precertificate, then the precertificate's SCT of type <spanx style="verb">precert_sct</spanx> would not be appropriate; instead, a fresh SCT of type <spanx style="verb">x509_sct</spanx> should be generated).
```
benlaurie: If you're going to make it a new sentence, I would drop the brackets.

rfc6962-bis.xml
```diff
@@ -462,13 +458,10 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
 
     struct {
         CMSPrecert pre_certificate;
-        ASN.1Cert precertificate_chain&lt;0..2^24-1&gt;;
+        ASN.1Cert precertificate_chain&lt;1..2^24-1&gt;;
```
benlaurie: Heh, I'm sure the lower bound is more than 1 :-)

robstradling: It's >0 because (at least) the root certificate MUST be included in precertificate_chain.  (The root certificate can't be the pre_certificate, and we've said that the root must be stored as part of the entry).

I think 1 is the correct lower bound.  Consider a precertificate for a name-constrained intermediate that will be signed directly by a root - surely the precertificate_chain will contain precisely one certificate (i.e. the root certificate)?

benlaurie: Isn't that number in bytes, though?

rfc6962-bis.xml
robstradling: The brackets are consistent with other parts of the doc, ISTM.  If you want to change that, how about doing it throughout the doc in a separate PR?

$ grep " (Note" rfc6962-bis.xml | wc -l
6

rfc6962-bis.xml
robstradling: No.  precertificate_chain is a vector of 1 or more ASN.1Certs, not 1 or more bytes.

robstradling: So the lower bound of ASN.1Cert could be increased (to some number >1).  I suggest not in this PR though.

benlaurie: Yeah, I wasn't really expecting you to make a change, just noting that 1 is only a teensy bit more correct than 0. :-)

rfc6962-bis.xml
```diff
@@ -1668,5 +1745,72 @@ value="Montreal"/>
       </reference>
 
     </references>
+
+    <section title="TransItemV1" anchor="TransItemV1">
+      <figure>
+        <preamble>
+          TODO: Finish writing this section. Or should it be in a separate document?
```
benlaurie: I think it should be in this document.

rfc6962-bis.xml
```diff
@@ -481,26 +474,119 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
-    } CertInfo;
-
+        SctExtension sct_extensions&lt;0..2^16-1&gt;;
+    } TimestampedCertificateEntryDataV2;</artwork>
+        </figure>
+        <t>
+          <spanx style="verb">timestamp</spanx> is the <xref target="RFC5905">NTP Time</xref> at which the certificate or precertificate was accepted by the log, measured in milliseconds since the epoch (January 1, 1970, 00:00), ignoring leap seconds.
+        </t>
+        <t>
+          <spanx style="verb">issuer_key_hash</spanx> is the HASH of the public key of the CA that issued the certificate or precertificate, calculated over the DER encoding of the key represented as SubjectPublicKeyInfo. This is needed to bind the CA to the certificate or precertificate, making it impossible for the corresponding SCT to be valid for any other certificate or precertificate whose TBSCertificate matches <spanx style="verb">tbs_certificate</spanx>.
```
benlaurie: Add a reference for SubjectPublicKeyInfo?

rfc6962-bis.xml
```diff
@@ -481,26 +474,119 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+        </t>
+        <t>
+          <spanx style="verb">tbs_certificate</spanx> is the DER encoded TBSCertificate from either the <spanx style="verb">leaf_certificate</spanx> (in the case of an <spanx style="verb">X509ChainEntry</spanx>) or the <spanx style="verb">pre_certificate</spanx> (in the case of a <spanx style="verb">PrecertChainEntryV2</spanx>). (Note that a precertificate's TBSCertificate can be reconstructed from the issued certificate's TBSCertificate by redacting the domain name labels indicated by the redacted labels extension, and deleting the SCT list extension and redacted labels extension).
```
benlaurie: Add a reference for TBSCertificate?

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+          <spanx style="verb">log_id</spanx> is this log's unique ID, encoded in an opaque vector as described in <xref target="log_id"/>.
         </t>
         <t>
-          <spanx style="verb">tbs_certificate</spanx> is the DER-encoded
-TBSCertificate component of the precertificate. Note that it is also possible to
-reconstruct this TBSCertificate from the issued certificate by extracting the
-TBSCertificate from it, redacting the domain name labels indicated by the
-redacted labels extension, and deleting the SCT list extension and redacted
-labels extension.
+          <spanx style="verb">timestamp</spanx> matches the timestamp from the <spanx style="verb">TimestampedCertificateEntryDataV2</spanx> structure encapsulated in the <spanx style="verb">timestamped_entry</spanx>.
```
benlaurie: s/matches/is equal to/?

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
         </t>
         <t>
-          <spanx style="verb">sct_extension_type</spanx> identifies a single extension from the IANA registry in <xref target="sct_extension_types"/>.
+          <spanx style="verb">sct_extension_type</spanx> identifies a single extension from the IANA registry in <xref target="sct_extension_types"/>. Currently, no extensions are specified.
```
benlaurie: s/Currently/At the time of writing/?

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
         </t>
         <t>
-          <spanx style="verb">entry_type</spanx> is the type of entry stored in <spanx style="verb">signed_entry</spanx>. New <spanx style="verb">LogEntryType</spanx> values may be added to <spanx style="verb">signed_entry</spanx> without increasing the <spanx style="verb">MerkleTreeLeaf</spanx> version. <xref target="client_messages"/> explains how clients should handle unknown entry types.
+          <spanx style="verb">tree_size</spanx> matches the tree size from the <spanx style="verb">TreeHeadDataV2</spanx> structure encapsulated in <spanx style="verb">merkle_tree_head</spanx>.
         </t>
         <t>
-          <spanx style="verb">signed_entry</spanx> is the <spanx style="verb">signed_entry</spanx> of the corresponding SCT.
+          <spanx style="verb">root_hash</spanx> matches the root hash from the <spanx style="verb">TreeHeadDataV2</spanx> structure encapsulated in <spanx style="verb">merkle_tree_head</spanx>.
```
benlaurie: s/matches/is equal to/? Or even just "is".

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
         </t>
         <t>
-          <spanx style="verb">extensions</spanx> are the <spanx style="verb">extensions</spanx> of the corresponding SCT.
+          <spanx style="verb">sth_extension_type</spanx> identifies a single extension from the IANA registry in <xref target="sth_extension_types"/>. Currently, no extensions are specified.
```
benlaurie: At the time of writing, again?

rfc6962-bis.xml
```diff
@@ -759,14 +826,14 @@ messages.
             <t hangText="Inputs:">
               <list style="hanging">
                 <t hangText="chain:">
-                  An array of base64 encoded certificates. The first element is the end-entity certificate; the second chains to the first and so on to the last, which is either the root certificate or a certificate that chains to a known root certificate.
+                  An array of base64 encoded certificates. The first element is the certificate for which the submitter desires an SCT; the second chains to the first and so on to the last, which is either an accepted root certificate or a certificate that chains to an accepted root certificate.
```
benlaurie: Do we need to defined "accepted root certificate"?

rfc6962-bis.xml
```diff
@@ -859,20 +930,20 @@ an accepted root certificate.
...
                 </t>
                 <t hangText="sth:">
-                  A base64 encoded SignedTreeHead.
+                  A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">signed_tree_head</spanx>, signed by this log.
                 </t>
               </list>
             </t>
             <t>
-              Note that no signature is required for the <spanx style="verb">consistency</spanx> output as it is used to verify <spanx style="verb">sth</spanx>, which is signed.
+              Note that no signature is required for the <spanx style="verb">consistency</spanx> output as it is used to verify STHs, which are signed.
```
benlaurie: if "the consistency output" is singular, than isn't "the STH" it verifies also singular?

rfc6962-bis.xml
```diff
@@ -906,23 +979,20 @@ an accepted root certificate.
...
             </t>
             <t hangText="Outputs:">
               <list style="hanging">
-                <t hangText="leaf_index:">
-                  The 0-based index of the entry corresponding to the <spanx style="verb">hash</spanx> parameter.
-                </t>
-                <t hangText="audit_path:">
-                  An array of base64 encoded Merkle Tree nodes proving the inclusion of the chosen certificate.
+                <t hangText="inclusion:">
+                  A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">inclusion_proof</spanx> whose <spanx style="verb">audit_path</spanx> array of Merkle Tree nodes proves the inclusion of the chosen certificate in the selected STH.
```
benlaurie: Hmm. It seems a little odd that we're using the terms "inclusion proof" and "audit path", which kinda mean the same thing.

rfc6962-bis.xml
```diff
@@ -906,23 +979,20 @@ an accepted root certificate.
...
                 </t>
                 <t hangText="sth:">
-                  A base64 encoded SignedTreeHead.
+                  A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">signed_tree_head</spanx>, signed by this log.
                 </t>
               </list>
             </t>
             <t>
-              Note that no signature is required for the <spanx style="verb">leaf_index</spanx> or <spanx style="verb">audit_path</spanx> outputs as they are used to verify inclusion in <spanx style="verb">sth</spanx>, which is signed.
+              Note that no signature is required for the <spanx style="verb">inclusion</spanx> output as it is used to verify inclusion in the selected STH, which is signed.
```
benlaurie: Note that this time the STH is singular :-)

rfc6962-bis.xml
robstradling: I agree that it's a minor point.  Nevertheless, there can never be 0 elements in the precertificate_chain, but there can be 1.  So 1 is correct.  :-)

rfc6962-bis.xml
robstradling: We already have...
  TBSCertificate <xref target="RFC5280"/>
...earlier in the document.  Is that not sufficient?

rfc6962-bis.xml
robstradling: My intended meaning was "s/verify STHs, which are signed/verify the consistency between two STHs, which are signed/".
Actually, I'm a bit uncomfortable with both "verify STHs" and "verify an STH", because they kinda give the impression that this verification is absolute (i.e. proving that the log isn't misbehaving), whereas in fact this verification is only relative (i.e. STH 2 is consistent with STH 1, but this doesn't prove that the log hasn't misbehaved 'cos STH 1 may or may not be consistent with all STHs that came before it).

rfc6962-bis.xml
robstradling: So, s/audit_path/inclusion_path/ then?

Then we'd have "consistency_proof" and "inclusion_proof" for the TransItem types; "consistency" and "inclusion" for the API outputs; and "consistency_path" and "inclusion_path" for the NodeHash vectors.

rfc6962-bis.xml
robstradling: Indeed.  So nothing to change here, right?

rfc6962-bis.xml
```diff
@@ -1231,18 +1293,22 @@ but it is expected there will be a variety.
...
         </t>
         <t>
-          A TLS client MAY audit the corresponding log by requesting, and
-          verifying, a Merkle audit proof for said certificate.
-          If the TLS client holds an STH that predates the SCT, it MAY, in the process of auditing, request a new STH from the log (<xref target="fetch_sth"/>), then verify it by requesting a consistency proof (<xref target="fetch_consistency"/>).
+          By validating SCTs and inclusion proofs, TLS clients can thus determine whether certificates are compliant. A certificate not accompanied by a valid SCT or inclusion proof MUST NOT be considered compliant by TLS clients. However, specifying the TLS clients' behavior once compliance or non-compliance has been determined (for example, whether a certificate should be rejected due to non-compliance) is outside the scope of this document.
+        </t>
+        <t>
+          If the TLS client holds an STH that predates the SCT, it MAY, in the process of auditing, request a new STH from the log (<xref target="get-sth"/>), then verify it by requesting a consistency proof (<xref target="get-sth-consistency"/>).
```
benlaurie: If the client uses get-all-by-hash, then it will already have the new STH.

rfc6962-bis.xml
robstradling: Let's do that in http://trac.tools.ietf.org/wg/trans/trac/ticket/102, as well as do "s/accepted root certificate/accepted trust anchor/".  I think we should add an "Accepted Trust Anchors" section just before the "Log Entries" section.

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/b6c13d3ad71def08e3d1fdefcb8bd885506f9c00

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/c42b1ea0f759c380708e5d0886dc726c127141a8

rfc6962-bis.xml
robstradling: https://github.com/robstradling/certificate-transparency-rfcs/commit/21af85aff597cdc2adecee6ac1b8516c4a2ceb54

rfc6962-bis.xml
robstradling: https://github.com/robstradling/certificate-transparency-rfcs/commit/21af85aff597cdc2adecee6ac1b8516c4a2ceb54

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/c42b1ea0f759c380708e5d0886dc726c127141a8

rfc6962-bis.xml
robstradling: I've done "s/verify STHs, which are signed/verify the consistency between two STHs, which are signed/" in https://github.com/robstradling/certificate-transparency-rfcs/commit/dcb9376f35cd866d4fece3d8a75931581ed5d589

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/f85a8ad72c01fdbe4377cd836ec0b66ef7ec8e47

rfc6962-bis.xml
robstradling: I've done "s/audit_path/inclusion_path/" in https://github.com/robstradling/certificate-transparency-rfcs/commit/eaf493164e09ed9eab84c3b1e4184dea52e374c3

eranmes: Per http://trac.tools.ietf.org/wg/trans/trac/ticket/16, shouldn't it be 'inclusion proof' ?

rfc6962-bis.xml
```diff
@@ -481,26 +474,119 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+          <spanx style="verb">data</spanx> is the encapsulated data structure. The various structures named with the <spanx style="verb">DataV2</spanx> suffix are defined in later sections of this document.
+        </t>
+        <t>
+          <spanx style="verb">item_extension_type</spanx> identifies a single extension from the IANA registry in <xref target="item_extension_types"/>.
+        </t>
+        <t>
+          The interpretation of the <spanx style="verb">item_extension_data</spanx> field is determined solely by the value of the <spanx style="verb">item_extension_type</spanx> field. Each document that registers a new <spanx style="verb">item_extension_type</spanx> must describe how to interpret the corresponding <spanx style="verb">item_extension_data</spanx>.
+        </t>
+        <t>
+          <spanx style="verb">item_extensions</spanx> is a vector of 0 or more item extensions. This vector MUST NOT include more than one extension with the same <spanx style="verb">item_extension_type</spanx>. The extensions in the vector MUST be ordered by the value of the <spanx style="verb">item_extension_type</spanx> field, smallest value first.
```
eranmes: (1) why must it not include more than one extension with the same type?
(2) Why should the extensions be ordered?

rfc6962-bis.xml
```diff
@@ -481,26 +474,119 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
     opaque TBSCertificate&lt;1..2^24-1&gt;;
 
     struct {
+        uint64 timestamp;
         opaque issuer_key_hash[HASH_SIZE];
         TBSCertificate tbs_certificate;
-    } CertInfo;
-
+        SctExtension sct_extensions&lt;0..2^16-1&gt;;
+    } TimestampedCertificateEntryDataV2;</artwork>
```
eranmes: nit: </artwork> in a new line?

rfc6962-bis.xml
```diff
@@ -481,26 +474,119 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+        </t>
+        <t>
+          <spanx style="verb">issuer_key_hash</spanx> is the HASH of the public key of the CA that issued the certificate or precertificate, calculated over the DER encoding of the key represented as <xref target="RFC5280">SubjectPublicKeyInfo</xref>. This is needed to bind the CA to the certificate or precertificate, making it impossible for the corresponding SCT to be valid for any other certificate or precertificate whose TBSCertificate matches <spanx style="verb">tbs_certificate</spanx>.
+        </t>
+        <t>
+          <spanx style="verb">tbs_certificate</spanx> is the DER encoded TBSCertificate from either the <spanx style="verb">leaf_certificate</spanx> (in the case of an <spanx style="verb">X509ChainEntry</spanx>) or the <spanx style="verb">pre_certificate</spanx> (in the case of a <spanx style="verb">PrecertChainEntryV2</spanx>). (Note that a precertificate's TBSCertificate can be reconstructed from the issued certificate's TBSCertificate by redacting the domain name labels indicated by the redacted labels extension, and deleting the SCT list extension and redacted labels extension).
```
eranmes: Is the note about reconstructing the precert's TBSCertificate detailed somewhere else?

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
-            uint64 timestamp;
-            LogEntryType entry_type;
-            select(entry_type) {
-                case x509_entry: CertInfo;
-                case precert_entry_V2: CertInfo;
-            } signed_entry;
-            SctExtensions extensions;
-        };
-    } SignedCertificateTimestamp;</artwork>
+            TransItem timestamped_entry;
```
eranmes: General comment: Seems like we're going to great length to accommodate V1 structures by introducing the TransItem -> TransItemV2 abstraction.
Why do you think it's necessary?

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+        SthExtension sth_extensions&lt;0..2^16-1&gt;;
+    } TreeHeadDataV2;</artwork>
+        </figure>
         <t>
-          <spanx style="verb">timestamp</spanx> is the current <xref target="RFC5905">NTP Time</xref>, measured since the epoch (January 1, 1970, 00:00), ignoring leap seconds, in milliseconds.
+          <spanx style="verb">timestamp</spanx> is the current <xref target="RFC5905">NTP Time</xref>, measured in milliseconds since the epoch (January 1, 1970, 00:00), ignoring leap seconds.
         </t>
         <t>
-          <spanx style="verb">entry_type</spanx> may be implicit from the context in which the SCT is presented.
+          <spanx style="verb">tree_size</spanx> equals the number of entries currently in the log's Merkle Tree.
```
eranmes: nit: s/equals/is/
The 'equals' below (in the SignedTreeHeadDataV2) makes this 'equals' somewhat confusing: 
I understand why you phrased the description of the fields in SignedTreeHeadDataV2 as you did, which is why I don't think it makes sense here.

rfc6962-bis.xml
```diff
@@ -510,186 +596,170 @@ Section 8.2 of <xref target="RFC3447"/>) using a key of at least 2048 bits.
...
+          <spanx style="verb">log_id</spanx> is this log's unique ID, encoded in an opaque vector as described in <xref target="log_id"/>.
+        </t>
+        <t>
+          <spanx style="verb">tree_size_1</spanx> is the size of the older tree.
+        </t>
+        <t>
+          <spanx style="verb">tree_size_2</spanx> is the size of the newer tree.
+        </t>
+        <t>
+          <spanx style="verb">consistency_path</spanx> is a vector of base64 encoded Merkle Tree nodes proving the consistency of two STHs.
```
eranmes: I would imagine base64 encoding is only necessary when transmitting data using JSON?
Furthermore, if it's base64-encoded, then wouldn't the field end up being bigger than HASH_SIZE?

rfc6962-bis.xml
```diff
@@ -1102,72 +1171,65 @@ entry specified by <spanx style="verb">start</spanx>.
...
 
     struct {
-        SerializedSCT sct_list&lt;1..2^16-1&gt;;
-    } SignedCertificateTimestampList;</artwork>
+        SerializedTransItem trans_item_list&lt;1..2^16-1&gt;;
+    } TransItemList;</artwork>
       </figure>
       <t>
-        Here, <spanx style="verb">SerializedSCT</spanx> is an opaque byte string that contains the serialized SCT structure. This encoding ensures that TLS clients can decode each SCT individually (i.e., if there is a version upgrade, out-of-date clients can still parse old SCTs while skipping over new SCTs whose versions they don't understand).
+        Here, <spanx style="verb">SerializedTransItem</spanx> is an opaque byte string that contains the serialized <spanx style="verb">TransItem</spanx> structure. This encoding ensures that TLS clients can decode each <spanx style="verb">TransItem</spanx> individually (so, for example, if there is a version upgrade, out-of-date clients can still parse old SCTs while skipping over new SCTs whose versions they don't understand).
```
eranmes: Nit: Still mentioning SCTs while the rest of the sentence refers to SerializedTransItems.

rfc6962-bis.xml
```diff
@@ -1102,72 +1171,65 @@ entry specified by <spanx style="verb">start</spanx>.
...
-            <artwork>
-    PrecertificateSCTList ::= OCTET STRING</artwork>
-          </figure>
+        <section title="Certificate Extension" anchor="cert_transinfo_extension">
           <t>
-            <spanx style="verb">PrecertificateSCTList</spanx> contains a <spanx style="verb">SignedCertificateTimestampList</spanx> whose SCTs all have the <spanx style="verb">precert_entry_V2</spanx> <spanx style="verb">LogEntryType</spanx>.
-          </t>
-          <t>
-            Upon receiving the certificate, clients can reconstruct the original <spanx style="verb">TBSCertificate</spanx> to verify the SCT signatures.
+            A certification authority may include a Transparency Information X.509v3 extension in a certificate. The included SCTs or inclusion proofs MUST be either for a precertificate that corresponds to this certificate, or for a name-constrained intermediate to which this certificate chains.
```
eranmes: Nit: A bit unclear if this sentence implies the only TransItems allowed to be embedded in a certificate are inclusion proofs  or SCTs.

rfc6962-bis.xml
```diff
@@ -1231,18 +1293,22 @@ but it is expected there will be a variety.
...
         </t>
         <t>
-          A TLS client MAY audit the corresponding log by requesting, and
-          verifying, a Merkle audit proof for said certificate.
-          If the TLS client holds an STH that predates the SCT, it MAY, in the process of auditing, request a new STH from the log (<xref target="fetch_sth"/>), then verify it by requesting a consistency proof (<xref target="fetch_consistency"/>).
+          By validating SCTs and inclusion proofs, TLS clients can thus determine whether certificates are compliant. A certificate not accompanied by a valid SCT or inclusion proof MUST NOT be considered compliant by TLS clients. However, specifying the TLS clients' behavior once compliance or non-compliance has been determined (for example, whether a certificate should be rejected due to non-compliance) is outside the scope of this document.
```
eranmes: Nit: Be more explicit about certificates accompanied by SCTs or inclusion proofs being considered compliant.

rfc6962-bis.xml
robstradling: If two different objects (the TransItem object and its member NodeHash vector) are both called "inclusion_proof", how will the reader know which of those objects we're talking about?
ISTM that it makes sense to give them different (but related) names, rather than have to qualify which object we're talking about every time we mention one of them!

rfc6962-bis.xml
robstradling: Because that's what Steve Kent wanted when we defined the top-level SCT extension syntax a few months ago.  ISTM that it made sense to define STH extensions and TransItem extensions in the same manner.
BTW, see http://www.ietf.org/mail-archive/web/trans/current/msg01388.html for some discussion of why Steve wanted the extensions to be ordered.

rfc6962-bis.xml
robstradling: Between ";" and "</artwork"?
The document consistently does _not_ include a newline there, so that "xml2rfc --html" doesn't add an unnecessary blank line at the bottom of each artwork.

rfc6962-bis.xml
robstradling: I think it's currently only detailed on that line.
Perhaps it should be moved to the "TLS Client" section later in the document?

rfc6962-bis.xml
robstradling: I wanted to ensure that the definition of TransItem left open the possibility that V1 structures could be accommodated.  It seemed like something that we might need, and I figured it was easier to take it out (if we decide we don't need it) than to try to bolt it on later.

Personally, I think the gain is worth the (complexity) pain.  RFC6962-compliant logs aren't going to go away overnight.  This mechanism lets us deliver V1 inclusion proofs.  It could be used for gossiping V1 structures.

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/5a0658ff5b674b30745c2021ebb8c44488582db4

rfc6962-bis.xml
robstradling: Good catch!

robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/d2e45ea8f7384ae823e021137d65c4dcd8863ba6

rfc6962-bis.xml
robstradling: I'd left it mentioning SCTs 'cos it was only a "for example", but I'm happy to change it.

Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/02f555c9e4ddfb0f7c233a8397cd93cc13897ad5

rfc6962-bis.xml
robstradling: How about I do: s/The/Any/ ?

I think we should leave open the possibility that in the future there might be other TransItem types that a CA might legitimately want to embed in a cert.

rfc6962-bis.xml
robstradling: Does the following text work for you?

"To be considered compliant, a certificate MUST be accompanied by at least one valid SCT or at least one valid inclusion proof. A certificate not accompanied by any valid SCTs or any valid inclusion proofs MUST NOT be considered compliant by TLS clients. However, specifying the..."

eranmes: Yes, thanks.

rfc6962-bis.xml
eranmes: SGTM, especially since there'd be a clarification that SCT or inclusion proof are needed for a cert to be considered compliant.

eranmes:
LGTM assuming the two changes (re when certificates are considered compliant and  s/The/Any/) are made.

rfc6962-bis.xml
robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/d6027da0a918e86ae20860b0aa54b38b8d3a3b7f

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/da8b93ff7c62cfffafe39fde9b59662c044d21fb

PR #90, opened 2015-11-03
Ticket 64: Remove specification of signature algorithms.

eranmes:
Similar to how hash algorithms are specified.


eranmes:
@benlaurie please take a look.

eranmes:
@robstradling FYI

PR #91, opened 2015-11-03
Ticket 78: More text about algorithm agility

eranmes:
The added text should make it clear why supporting algorithm agility (in practice, allowing logs to support multiple signature and hash algorithms) is too complex and unnecessary given the simpler alternative of freezing a log.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1460,7 +1460,14 @@ but it is expected there will be a variety.
 
     <section title="Algorithm Agility">
       <t>
-        It is not possible for a log to change any of its algorithms part way through its lifetime. If it should become necessary to deprecate an algorithm used by a live log, then the log should be frozen as specified in <xref target="metadata"/> and a new log should be started. If necessary, the new log can contain existing entries from the frozen log, which monitors can verify are an exact match.
+        It is not possible for a log to change any of its algorithms part way through its lifetime:
+        <list style="hanging">
+          <t hangText="Signature algorithm:">SCT signatures must remain valid so signature algorithms can only be added, not removed.</t>
+          <t hangText="Hash algorithm:">A log would have to support the old and new hash algorithms to allow backwards-compatibility with clients that are not aware of a hash algorithm change. </t>
+        </list>
+        Allowing multiple signature or hash algorithms for a log would require that all data structures support it and would significantly complicate client implementation, which is why it is not supported by this document.
+      </t>
+      <t>If it should become necessary to deprecate an algorithm used by a live log, then the log should be frozen as specified in <xref target="metadata"/> and a new log should be started. Certificates in the frozen log that have not yet expired and require new SCTs should be submitted to the new log and the SCTs from that log used instead.
       </t>
     </section>
 


eranmes:
@benlaurie PTAL.

eranmes:
@robstradling FYI

robstradling:
LGTM.  I think this only addresses some of Steve Kent's concerns though (http://trac.tools.ietf.org/wg/trans/trac/ticket/78#comment:3).

benlaurie:
Not keen on this change - it doesn't really add anything useful, IMO. As Rob says, it also doesn't address the concern. Not convinced, in any case, that the concern is really addressable.

eranmes:
@benlaurie PTAL.

rfc6962-bis.xml
```diff
@@ -1467,7 +1467,14 @@ but it is expected there will be a variety.
...
+        It is not possible for a log to change any of its algorithms part way through its lifetime:
+        <list style="hanging">
+          <t hangText="Signature algorithm:">SCT signatures must remain valid so signature algorithms can only be added, not removed.</t>
+          <t hangText="Hash algorithm:">A log would have to support the old and new hash algorithms to allow backwards-compatibility with clients that are not aware of a hash algorithm change. </t>
+        </list>
+        Allowing multiple signature or hash algorithms for a log would require that all data structures support it and significantly complicate client implementation, which is why it is not supported by this document.
+      </t>
+      <t>If it should become necessary to deprecate an algorithm used by a live log, then the log should be frozen as specified in <xref target="metadata"/> and a new log should be started. Certificates in the frozen log that have not yet expired and require valid SCTs should be re-logged in the new log and the SCTs from that log used instead.
```
benlaurie: Suggest "require new SCTs" rather than valid (old SCTs may not be invalid).

benlaurie:
One minor correction and then LGTM.

Martin2112:
I think my PR covers the same thing. This version looks better.

rfc6962-bis.xml
```diff
@@ -1467,7 +1467,14 @@ but it is expected there will be a variety.
 
     <section title="Algorithm Agility">
       <t>
-        It is not possible for a log to change any of its algorithms part way through its lifetime. If it should become necessary to deprecate an algorithm used by a live log, then the log should be frozen as specified in <xref target="metadata"/> and a new log should be started. If necessary, the new log can contain existing entries from the frozen log, which monitors can verify are an exact match.
+        It is not possible for a log to change any of its algorithms part way through its lifetime:
+        <list style="hanging">
+          <t hangText="Signature algorithm:">SCT signatures must remain valid so signature algorithms can only be added, not removed.</t>
+          <t hangText="Hash algorithm:">A log would have to support the old and new hash algorithms to allow backwards-compatibility with clients that are not aware of a hash algorithm change. </t>
+        </list>
+        Allowing multiple signature or hash algorithms for a log would require that all data structures support it and significantly complicate client implementation, which is why it is not supported by this document.
```
robstradling: For clarity:
s/and significantly/and would significantly/

We don't intend to require the data structures to significantly complicate client implementation!

rfc6962-bis.xml
robstradling: s/re-logged in/submitted to/

We already talk about submitting certs to logs.  There's no need to introduce a new verb that is intended to mean the same thing.

eranmes:

PR #93, opened 2015-11-27
Ticket 83: Explain the necessity of deterministic signatures.

eranmes:


ln5:
I suggest s/can potentially leak the secret material/can potentially expose the secret material/1.

benlaurie:
LGTM with or without Linus' change.

robstradling:
LGTM too.

PR #94, opened 2015-11-27
Ticket 95: Recommend limit for get-entries response size.

eranmes:


rfc6962-bis.xml
```diff
@@ -1130,7 +1130,7 @@ treating unrecognized leaves as opaque input to the tree.
         </t>
         <t>
           Logs MAY restrict the number of entries that can be retrieved per
-<spanx style="verb">get-entries</spanx> request.  If a client requests more
+<spanx style="verb">get-entries</spanx> request (the limit of 1,000 entries is suggested).  If a client requests more
```
benlaurie: Why are we suggesting a limit? Also "_a_ limit..."

benlaurie:
My position on ticket 95 hasn't changed. I don't see any use in this. Instead propose we close ticket 95.

robstradling:
For crt.sh's, I currently fetch (up to) 1,024 entries per get-entries request.  I arrived at this strategy by trial and error.  It would be nice to have more determinism here, but simply suggesting a limit in the doc wouldn't actually help with that, because I'd still have to figure out what the logs are actually doing by trial and error.  So I agree with Ben: let's close ticket 95 as a WONTFIX.

eranmes:
Abandoning this PR as suggested by Ben and Rob.

PR #96, opened 2015-11-27
Don't call logs "untrusted".

benlaurie:
@eranmes @robstradling

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -79,10 +79,10 @@ submissions and queries that are defined in this document.
     <section title="Introduction">
       <t>
         Certificate transparency aims to mitigate the problem of misissued
-certificates by providing publicly auditable, append-only, untrusted logs of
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -79,10 +79,10 @@ submissions and queries that are defined in this document.
     <section title="Introduction">
       <t>
         Certificate transparency aims to mitigate the problem of misissued
-certificates by providing publicly auditable, append-only, untrusted logs of
-all issued certificates. The logs are publicly auditable so that it is possible
-for anyone to verify the correctness of each log and to monitor when new
-certificates are added to it. The logs do not themselves prevent misissue, but
+certificates by providing append-only logs of all issued certificates. The logs
```
robstradling: This suggested revision didn't reach the issue tracker:
http://www.ietf.org/mail-archive/web/trans/current/msg01692.html

s/all issued certificates/issued certificates/

PR #99, opened 2015-12-21
Ticket 114: Name the OCSP Stapling TLS extension correctly.

robstradling:


rfc6962-bis.xml
```diff
@@ -1175,7 +1175,7 @@ entry specified by <spanx style="verb">start</spanx>.
...
           </t>
           <t>
-            An <xref target='RFC6960'>Online Certificate Status Protocol (OCSP)</xref> response extension (see <xref target="ocsp_transinfo_extension"/>), where the OCSP response is provided in the <spanx style="verb">certificate_status</spanx> TLS extension (Section 8 of <xref target='RFC6066'/>), also known as OCSP stapling. This mechanism is already widely (but not universally) implemented. It also allows SCTs and inclusion proofs to be updated on the fly.
+            An <xref target='RFC6960'>Online Certificate Status Protocol (OCSP)</xref> response extension (see <xref target="ocsp_transinfo_extension"/>), where the OCSP response is provided in the <spanx style="verb">CertificateStatus</spanx> message, provided that the TLS client included the <spanx style="verb">status_request</spanx> extension in the (extended) <spanx style="verb">ClientHello</spanx> (Section 8 of <xref target='RFC6066'/>). This mechanism, popularly known as OCSP stapling, is already widely (but not universally) implemented. It also allows SCTs and inclusion proofs to be updated on the fly.
```
eranmes: Nit: Do you think it's worth adding 'by the CA' to the end of the last sentence?

eranmes:
Only one minor comment, LGTM otherwise.

rfc6962-bis.xml
robstradling: It requires the cooperation of the CA, but it also requires the server to fetch the updated OCSP response "on the fly" and to start stapling it in new handshakes.  Since the CA can't force the server to fetch the updated OCSP response, I think I'd prefer to append "with the cooperation of the CA".  Does that work for you?

eranmes: Yes.

robstradling: OK, I'll leave it out and merge it as is.

PR #100, opened 2016-01-12
Fix the consistency checking proof to account for sizes.

eranmes:
Without checking sn, the same proof would be accepted for multiple tree sizes.

Thanks to David Benjamin for finding this out.


benlaurie:
You should also update https://github.com/google/certificate-transparency-rfcs/blob/master/test_rfc_algorithms.py - ideally after adding tests that reveal the bugs...

eranmes:
Done, PTAL.

PR #101, opened 2016-01-19
Ticket 102: "root" should be "trust anchor".

robstradling:


benlaurie:
LGTM, but just want to check that there's really nothing dumb about specifying it this way? Just wondering why no-one pointed it out before I did...

robstradling:
Not dumb at all.  RFC5280 uses the term "trust anchor" a lot, but never uses the term "root certificate".  And as you pointed out, some trust stores include intermediates as well as self-signed root certificates.  e.g. NSS includes this one: https://crt.sh/?id=156583

robstradling:
My guess is that no-one pointed it out before you did because: (i) relatively few people have actually read RFC5280 in detail and (ii) almost all trust anchors are self-signed root certificates and so (iii) no-one who needed this issue with the CT spec to be fixed was actually aware that this issue existed.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/56f42d9c48526798beb848af5c0e7cf1c528f470

PR #107, opened 2016-02-01
Address issue 132.

Martin2112:
I propose that we remove the confusing text as this isn't something we can cover in a single sentence and it doesn't seem that helpful to implementers.

Additionally, attempts to be prescriptive about operational events like this are likely to be incomplete and create uncertainty for log operators.

Please review and advise me if if there are things that logs are required to do in these circumstances that are not covered in the referenced section 9.1.


benlaurie:
There is already text in #91 modifying this...

PR #109, opened 2016-02-01
Ticket 109: Describing log shutdown.

eranmes:
Per dkg's request, describing what a responsible log shutdown
should look like by suggesting several actions to take.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -754,6 +754,17 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
           <spanx style="verb">inclusion_path</spanx> is a vector of Merkle Tree nodes proving the inclusion of the chosen certificate or precertificate.
         </t>
       </section>
+      <section title="Shutting down a log">
+        <t>
+          Log operators may decide to shut down a log for various reasons, such as deprecation of the signature algorithm. If there are entries in the log for certificates that have not yet expired, simply making TLS clients stop recognizing that log will have the effect of invalidating SCTs from that log. To avoid that, the following actions are suggested:
+          <list style="symbols">
+            <t>Make it known to clients and monitors that the log will be frozen.</t>
+            <t>Stop accepting new submissions (the error code "shutdown" should be returned for such requests).</t>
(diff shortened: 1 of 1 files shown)


eranmes:
@benlaurie  PTAL as you have concrete ideas on log shutdown.

rfc6962-bis.xml
```diff
@@ -752,6 +752,17 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
           <spanx style="verb">inclusion_path</spanx> is a vector of Merkle Tree nodes proving the inclusion of the chosen certificate or precertificate.
         </t>
       </section>
+      <section title="Shutting down a log">
+        <t>
+          Log operators may decide to shut down a log for various reasons, such as depracation of signature algorithm. If there are entries in the log for certificates that have not yet expired, simply making TLS clients stop recognizing that log will have the effect of invalidating SCTs from that log. To avoid that, the following actions are suggested:
```
benlaurie: s/depracation/deprecation/.

benlaurie: of _the_ signature algorithm

rfc6962-bis.xml
```diff
@@ -752,6 +752,17 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          <list style="symbols">
+            <t>Make it known to clients and monitors that the log will be frozen.</t>
+            <t>Stop accepting new submissions (the error code shutdown should be returned for such requests).</t>
```
benlaurie: the error code "shutdown", presumably

rfc6962-bis.xml
```diff
@@ -752,6 +752,17 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+            <t>Once MMD from the last accepted submission has passed and all pending submissions are incorporated, issue a final STH and publish it as a part of the log's metadata. At this point the log's private key is no longer needed and can be destroyed.</t>
```
benlaurie: Once _the_ MMD. But why wait for the MMD? Surely once the final STH is issued, you're done?

eranmes: That should guarantee that no client sees an SCT where (timestamp + MMD) > timestamp in final STH.
This is an attempt to simplify client implementation by not requiring it to be aware of the particular log shutdown case: a client can simply get the latest STH, find out that timestamp from SCT + MMD has passed and request an inclusion proof like any other case.

eranmes:
Addressed your comments, PTAL.

eranmes:
Per offline discussion, added a clarifying note. I'm sure there's better wording for saying the same.

PR #113, opened 2016-02-01
Add a few examples for when pre-cert SCT do not go in the certificate.

pphaneuf:
To address [ticket 127](https://tools.ietf.org/wg/trans/trac/ticket/127).


rfc6962-bis.xml
```diff
@@ -310,7 +310,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       </section>
       <section title="Precertificates" anchor="Precertificates">
         <t>
-          Alternatively, (root as well as intermediate) CAs may preannounce a certificate prior to issuance by <xref target="add-pre-chain">submitting a precertificate</xref> that the log can use to create an entry that will be valid against the issued certificate. The CA MAY incorporate the returned SCT in the issued certificate.
+          Alternatively, (root as well as intermediate) CAs may preannounce a certificate prior to issuance by <xref target="add-pre-chain">submitting a precertificate</xref> that the log can use to create an entry that will be valid against the issued certificate. The CA MAY incorporate the returned SCT in the issued certificate. Examples of situations where the returned SCT is not incorporated into the issued certificate would be in the case of domain name redaction, when an SCT is successfully generated but a network error prevents the CA from obtaining it, or the CA is attempting to keep the size of the certificate small (possibly using another mechanism to publish the SCT, such as an <xref target="ocsp_transinfo_extension">OCSP response</xref>).
```
robstradling: In "the case of domain name redaction", precertificates SCTs can be embedded in certs, or omitted.  But I think your text seems to wrongly imply that they're always omitted in this case.
Domain name redaction relies on precertificates, but we don't want to prevent the use of OCSP Stapling and transparency_info in conjunction with domain name redaction.
If domain name redaction is not being used, then your third example (CA...size...small...OCSP response) doesn't need to use precertificates at all.  i.e. a certificate SCT can be used instead.

So I think you need to condense your 3 examples down to 2...
1) your point about network errors.
2) domain name redaction combined with the desire to gain the benefits of OCSP stapling / transparency_info (as discussed in Section 7).

pphaneuf:
PTAL

rfc6962-bis.xml
```diff
@@ -310,7 +310,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       </section>
       <section title="Precertificates" anchor="Precertificates">
         <t>
-          Alternatively, (root as well as intermediate) CAs may preannounce a certificate prior to issuance by <xref target="add-pre-chain">submitting a precertificate</xref> that the log can use to create an entry that will be valid against the issued certificate. The CA MAY incorporate the returned SCT in the issued certificate.
+          Alternatively, (root as well as intermediate) CAs may preannounce a certificate prior to issuance by <xref target="add-pre-chain">submitting a precertificate</xref> that the log can use to create an entry that will be valid against the issued certificate. The CA MAY incorporate the returned SCT in the issued certificate. Examples of situations where the returned SCT is not incorporated into the issued certificate would be when an SCT is successfully generated but a network error prevents the CA from obtaining it, or the CA is using domain name redaction and intends to use another mechanism to publish SCTs (such as an <xref target="ocsp_transinfo_extension">OCSP response</xref> or the <xref target="tls_transinfo_extension">TLS extension</xref>).
```
benlaurie: The network error is a bit forced. How about instead: "when a CA sends the precertificate to multiple  logs, but only incorporates the SCTs that are returned first".

pphaneuf:
PTAAL

PR #114, opened 2016-02-01
Ticket 120: Clarify this is about the protocol.

eranmes:
It should be clear that certificate submission is not limited to one
entity in the protocol. Whether submitted certificates are accepted
by a log or not is up to the log's policy.


robstradling:
Please do s/109/120/ on the commit message when you merge.

eranmes:
Done.

PR #117, opened 2016-02-01
Ticket 141 auditing description

phad:
Remove the vague terminology 'partial information' in section 'Auditing'. Flesh out descriptions of the duties of auditing via STH and auditing via SCT.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1359,7 +1359,13 @@ but it is expected there will be a variety.
       </section>
       <section title="Auditing">
         <t>
-          Auditing is taking partial information about a log as input and verifying that this information is consistent with other partial information held. All clients described above may perform auditing as an additional function. The action taken by the client if audit fails is not specified, but note that in general if audit fails, the client is in possession of signed proof of the log's misbehavior.
+          Auditing ensures that the current published state of a log is reachable from previously published states that are known to be good, and that the promises made by the log in the form of SCTs have been kept.  Audits are performed by monitors or TLS clients.
+        </t>
+        <t>
+          A benign, conformant log publishes a series of STHs over time, each derived from the previous STH and the submitted entries incorporated into the log since publication of the previous STH.  This can be proven through auditing of STHs.  SCTs returned to TLS clients can be audited by verifying against the accompanying certificate, and using Merkle Inclusion Proofs, against the log's Merkle tree.
+        </t>
+        <t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1359,7 +1359,14 @@ but it is expected there will be a variety.
       </section>
       <section title="Auditing">
         <t>
-          Auditing is taking partial information about a log as input and verifying that this information is consistent with other partial information held. All clients described above may perform auditing as an additional function. The action taken by the client if audit fails is not specified, but note that in general if audit fails, the client is in possession of signed proof of the log's misbehavior.
+          Auditing ensures that the current published state of a log is reachable from previously published
+          states that are known to be good.  Audits are performed by monitors or TLS clients.
```
benlaurie: Why the line break?

phad: Probably just coding habits and the 80-col margin.
Should I just allow my editor to create long lines for RFC work?

benlaurie: Also, this is not really an accurate statement (e.g. it does not cover the inclusion proof case).

Maybe instead say "Auditing ensures that the information held by a client is consistent with the current published state of the log"?

rfc6962-bis.xml
```diff
@@ -1359,7 +1359,14 @@ but it is expected there will be a variety.
...
+        </t>
+        <t>
+          A benign, conformant log publishes a series of STHs over time, each derived from the previous STH and the submitted entries incorporated into the log since publication of the previous STH.  This can be proven through auditing of STHs.  SCTs returned to TLS clients can be audited by verifying against the accompanying certificate, and using Merkle Inclusion Proofs, against the log's Merkle tree.
+        </t>
+        <t>
+          The action taken by the auditor if an audit fails is not specified, but note that in general if audit fails, the auditor is in possession of signed proof of the log's misbehavior.
```
benlaurie: And if you're going to spell this out, you need to also spell out SCTs as mentioned below.

phad: Which bit do you mean?

rfc6962-bis.xml
phad: How about:  "Auditing ensures that the current published state of a log is reachable from previously published states that are known to be good, and that the promises made by the log in the form of SCTs have been kept.  Audits are performed by monitors or TLS clients."

It's more verbose but provides specifics.  The original objection was about using the term 'partial information' which was too woolly.

PR #118, opened 2016-02-01
Say that CAs can require CT by adding a TLS Feature extension.

benlaurie:


rfc6962-bis.xml
```diff
@@ -1203,6 +1203,14 @@ entry specified by <spanx style="verb">start</spanx>.
         </t>
       </section>
     </section>
+    <section title="TLS Feature Extension">
+      <t>
+	A CA MAY include the CT TLS extension number in a <xref target="RFC7633">TLS Feature Extension</xref> in root or intermediate certificates. This indicates that CT compliance is required for all certificates which are directly or indirectly issued by such certificates.
```
robstradling: I suggest: s/CT TLS extension/transparency_info TLS extension/
Might be worth xref'ing to the section above too.

robstradling: Also, it would be good to say something about when/why to (or not to) include the transparency_info TLS extension number in the TLS Feature extension in end-entity certificates.

rfc6962-bis.xml
```diff
@@ -1203,6 +1203,14 @@ entry specified by <spanx style="verb">start</spanx>.
...
+      </t>
+      <t>
+	Clients MUST treat certificates which fail to comply with this requirement as invalid.
```
robstradling: s/Clients/TLS clients/

rfc6962-bis.xml
benlaurie: I commented on that in the trac bug.

rfc6962-bis.xml
benlaurie: The reason I didn't say TLS clients is that you can process a certificate chain without being a TLS client...

robstradling: Good point.  A monitor could spot failures to comply with this requirement, for example.

robstradling: "A CA MAY include" doesn't belong in a section on TLS Servers.

"Clients MUST" doesn't belong in a section on TLS Servers.

robstradling: I suggest: move the "A CA MAY include" sentence into a new Section 8.2; and move the "Clients MUST" sentence into a new subsection of Section 9, somewhere before Section 9.2 (TLS Client).

benlaurie: This is not in the section on servers, it is a new section all on its own (it becomes section 8).

robstradling: Ah, I see.  But isn't it better to have all of the CA requirements in the CA section, and all of the Client requirements in the Client section (as per the current version of the draft)?

rfc6962-bis.xml
```diff
@@ -1226,6 +1226,11 @@ entry specified by <spanx style="verb">start</spanx>.
           </t>
         </section>
       </section>
+      <section title="TLS Feature Extension">
+	<t>
+	  A CA MAY include the <xref target="tls_transinfo_extension">transparency_info</xref> extension number in a <xref target="RFC7633">TLS Feature Extension</xref> in root or intermediate certificates. This indicates that CT compliance is required for all certificates which are directly or indirectly issued by such certificates.
+	</t>
+      </section>
```
robstradling: LGTM, but I'd still like to see text explaining when/why to (or not to) include the "transparency_info" extension number in the TLS Feature extension in end-entity certificates.  Two possible options (and there may be others): 1) CAs SHOULD NOT do that; if they do, clients should ignore it; 2) CAs MAY do that; if they do, it means that clients MUST reject the cert if it is non-compliant with 6962-bis.

rfc6962-bis.xml
```diff
@@ -1458,6 +1463,14 @@ but it is expected there will be a variety.
           </t>
         </section>
       </section>
+      <section title="TLS Feature Extension">
+	<t>
+	  If a certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> extension number in a <xref target="RFC7633">TLS Feature Extension</xref>, then CT compliance is required for all certificates which are directly or indirectly issued by that certificate.
+	</t>
+	<t>
+	  Clients MUST treat certificates which fail to comply with this requirement as invalid.
```
eranmes: nit: In other places in the document, certificates are considered 'compliant'/'non-compliant' with the RFC rather than valid/invalid.

rfc6962-bis.xml
benlaurie: Suggest you propose your own PR for that...

benlaurie:
PTAL

rfc6962-bis.xml
eranmes: Should have been clearer: What I suggest is you use the term 'non-compliant' rather than 'invalid'.

robstradling:
I kinda liked "invalid" in this particular case, given that it's about compliance with both CT and TLS Feature, not just compliance with CT.

However, since RFC7633 only goes as far as saying "Clients MAY refuse to accept the connection if the server does not accept a request for a specified feature" (rather than MUST), maybe "non-compliant" is better.

rfc6962-bis.xml
robstradling: OK, will do.  Do you have any preference on which option I should pick?

(Meanwhile, this PR LGTM).

rfc6962-bis.xml
benlaurie: I did, you are looking at an old diff. :-)

benlaurie:
Actually 7633 says "In the case that use of TLS with a valid certificate is mandated by
   explicit security policy, application protocol specification, or
   other means, the client MUST refuse the connection.  If the use of
   TLS with a valid certificate is optional, a client MAY accept the
   connection but MUST NOT treat the certificate as valid."

Anyway, I think you should go for the second version (and make it clear that it can use any of the mechanisms to be compliant, not just the TLS extension).

PR #119, opened 2016-02-01
Ticket 143:  re-organization of sections related to Multiple SCTs.

phad:
Text from section 12.6 "Multiple SCTs" is moved into section 7 where Multiple SCTs are first discussed.  The former section 12.6 is retained with a reference back to sect. 7.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1177,26 +1177,39 @@ entry specified by <spanx style="verb">start</spanx>.
           </t>
         </list>
       </t>
-      <t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1603,18 +1616,7 @@ certificates around the SCT timestamp.
       </section>
       <section title="Multiple SCTs">
```
benlaurie: You now have two sections with the same title..

phad: I don't think that's a problem, is it? I can link to one using a named
anchor.

benlaurie: I guess not, it just looked a bit weird.

PR #122, opened 2016-02-02
Add language on non-compliance handling inspired by RFC 7633.

benlaurie:


rfc6962-bis.xml
```diff
@@ -1477,6 +1477,21 @@ but it is expected there will be a variety.
 	  Clients MUST treat certificates which fail this requirement as non-compliant.
 	</t>
       </section>
+      <section title="Handling of Non-compliance">
+	<t>
+	  If the server presents a certificate chain that is non-compliant, there are two possibilites.
```
robstradling: s/server/TLS server/

Log servers can "present" certificate chains via get-entries. ;-)

robstradling: s/possibilites/possibilities/

PR #123, opened 2016-02-02
Ticket 113: Further advice about the TLS Feature extension.

robstradling:


rfc6962-bis.xml
```diff
@@ -1470,12 +1470,12 @@ but it is expected there will be a variety.
...
-	  If a certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> extension number in a <xref target="RFC7633">TLS Feature Extension</xref>, then CT compliance is required for all certificates which are directly or indirectly issued by that certificate.
-	</t>
-	<t>
-	  Clients MUST treat certificates which fail this requirement as non-compliant.
-	</t>
+        <t>
+          If any certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension, then CT compliance (using any of the mechanisms from <xref target="tls_servers"/>) is required.
+        </t>
+        <t>
+          Clients MUST treat certificates which fail this requirement as non-compliant.
```
robstradling: I think we discussed "Clients MUST" yesterday.  You pointed out that it's not just TLS clients that evaluate certificate chains.  However...
How would any other type of client evaluate whether or not CT compliance has been achieved?
It's only a TLS client that will receive SCTs / inclusion proofs that are sent via the transparency_info TLS extension, for example.

rfc6962-bis.xml
```diff
@@ -1226,10 +1226,10 @@ entry specified by <spanx style="verb">start</spanx>.
           </t>
         </section>
       </section>
-      <section title="TLS Feature Extension">
-	<t>
-	  A CA MAY include the <xref target="tls_transinfo_extension">transparency_info</xref> extension number in a <xref target="RFC7633">TLS Feature Extension</xref> in root or intermediate certificates. This indicates that CT compliance is required for all certificates which are directly or indirectly issued by such certificates.
-	</t>
+      <section title="TLS Feature Extension" anchor="tls_feature_extension">
+        <t>
+          A certification authority MAY include the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension in root, intermediate and end-entity certificates. When a certificate chain includes such a certificate, this indicates that CT compliance is required.
```
robstradling: BTW, I changed it to "TLS extension identifier" because that's the language used by RFC7633.  However, I wonder if "TLS ExtensionType value" would be more accurate (see http://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml).

rfc6962-bis.xml
benlaurie: Maybe say TLS client explicitly?

robstradling: Addressed in https://github.com/robstradling/certificate-transparency-rfcs/commit/e23f484dce5dd8fe774e83029c1d17583d786c7b

robstradling: With that change, I think this section should now be moved to become a subsection of the "TLS Client" section (currently numbered 9.2).

robstradling: I've done that in https://github.com/robstradling/certificate-transparency-rfcs/commit/b067f25730fa2a7ee544eae7f684cc41b983d5f0

robstradling:
PTAL

rfc6962-bis.xml
```diff
@@ -1308,6 +1308,14 @@ but it is expected there will be a variety.
         <t>
           If the TLS client holds an STH that predates the SCT, it MAY, in the process of auditing, request a new STH from the log (<xref target="get-sth"/>), then verify it by requesting a consistency proof (<xref target="get-sth-consistency"/>). Note that if the TLS client uses <spanx style="verb">get-all-by-hash</spanx>, then it will already have the new STH.
         </t>
+        <section title="TLS Feature Extension">
+          <t>
+            If any certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension, then CT compliance (using any of the mechanisms from <xref target="tls_servers"/>) is required.
```
benlaurie: This seems a little less clear than the original language - is compliance required for every cert in the chain? Just the EE cert? All certs from the marked cert downward? (I prefer that answer, btw).

robstradling: I reworded it deliberately so that those questions didn't need to be asked.  ;-)

A TLS client evaluates whether or not "compliance" is required on a per-chain basis rather than a per-certificate basis, I think.  (A TLS server doesn't have to supply SCTs for every cert in the chain, so the notion of requiring each cert in a chain to individually be "compliant" makes no sense to me).

With my language, "compliance" is required when a TLS client encounters a chain that contains any cert(s) that have the transparency_info TLS extension indentifier in the TLS Feature extension.

We define "compliance" elsewhere.  Section 9.2, for example, says:
 "To be considered compliant, a certificate MUST be accompanied by at
   least one valid SCT or at least one valid inclusion proof."

PR #128, opened 2016-02-18
Ticket #81: OIDs and IANA Considerations

robstradling:
Not sure if the "Log ID Registry" should be a subsection of "IANA Considerations".  I guess it depends on who's going to run this registry.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -321,7 +321,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
               It MUST be DER encoded.
             </t>
             <t>
-              <spanx style="verb">SignedData.encapContentInfo.eContentType</spanx> MUST be the OID &lt;TBD&gt;.
+              <spanx style="verb">SignedData.encapContentInfo.eContentType</spanx> MUST be the OID 1.3.101.78.
             </t>
             <t>
               <spanx style="verb">SignedData.encapContentInfo.eContent</spanx> MUST contain a TBSCertificate <xref target="RFC5280"/>, which MAY redact certain domain name labels that will be present in the issued certificate (see <xref target="redacting_subdomains"/>) and MUST NOT contain any SCTs, but which will be otherwise identical to the TBSCertificate in the issued certificate.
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1579,6 +1579,25 @@ but it is expected there will be a variety.
...
+        <t>
+          No action from IANA is necessary with regard to any of the OIDs in this document.
+        </t>
+        <section title="Log ID Registry" anchor="log_id_registry">
+          <t>
+            Symantec have delegated the 1.3.101.80 arc to the TRANS Working Group. From this arc, Log IDs will be allocated.
+          </t>
+          <t>
+            Additionally, Symantec have reserved all OIDs in the range from 1.3.101.8192 to 1.3.101.16383 to be used as Log IDs.
+          </t>
```
benlaurie: This is slightly confusing ... obvs the .80 arc is infinitely expandable. Why is there a second range? (Too lazy to work out lengths, is it because the second range is actually shorter?).

rfc6962-bis.xml
```diff
@@ -1579,6 +1579,25 @@ but it is expected there will be a variety.
...
+          <t>
+            TBD: policy for adding to the registry. And who's going to run the registry?
```
benlaurie: IANA should run the registry. Policy ... dunno ... ask the WG? Probably simply "expert review" would suffice?

rfc6962-bis.xml
robstradling: See http://trac.tools.ietf.org/wg/trans/trac/ticket/105#comment:10.
The .80 arc only provides 128 4-byte LogIDs, but it's infinitely expandable using >4-byte encodings.
The "second range" is a finite resource of 8,192 4-byte LogIDs.

rfc6962-bis.xml
robstradling: We opted to use OIDs for LogIDs so that IANA didn't have to be involved.  If IANA do have to be involved, then why don't we ditch OIDs and instead let LogIDs be 32-bit integers allocated by IANA?
(OK, so currently a log provider is free to allocate an OID from their own arc, but they probably had to obtain that arc from IANA anyway).

robstradling: Or is the freedom to allocate a LogID from a potentially non-IANA-allocated OID arc a sufficiently good reason to stick with OIDs?

rfc6962-bis.xml
benlaurie: OK.

rfc6962-bis.xml
benlaurie: Right, OIDs essentially allow anyone to bypass the register, so I see no issue with having it run by IANA for those who can't easily get OIDs. Also don't know who else would do it.

rfc6962-bis.xml
```diff
@@ -470,7 +470,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Log ID" anchor="log_id">
         <figure>
           <preamble>
-            Each log's operator allocates an OID for the purpose of uniquely identifying that log. This OID is specified in the log's metadata. Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
+            Each log is uniquely identified by an OID. A log's operator MUST either allocate the OID themselves (from an arc that belongs to them) or request an OID from one of the two Log ID Registries (see <xref target="log_id_registry1"/> and <xref target="log_id_registry2"/>). The OID is specified in the log's metadata. Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
```
benlaurie: Leave out "from an arc that belongs to them", since that is not actually a requirement.

benlaurie:
Apart from my comment, LGTM.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/0bbcfa5a6feb92cd781978d3757f251aaa8cf0bc

PR #131, opened 2016-03-08
Ticket 144: Overly redacted certs: Remove requirement from CAs

eranmes:
Also point to PSL as an example list of domains for which redacted
precerts should not be issued.


rfc6962-bis.xml
```diff
@@ -1602,13 +1602,7 @@ corrective action when a misissue is detected.
...
-certificate whose corresponding precertificate would be overly redacted, perhaps using the same mechanism for determining whether a wildcard in a domain name of a certificate is too broad.
+          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). CAs should take care not to redact domain name labels in precertificates too broadly, such that the entirety of the domain space below the unredacted part of the domain name may be controlled by multiple entities, such as domains in the <xref target="Public.Suffix.List">Public Suffix List</xref> (e.g. <spanx style="verb">?.com</spanx> and <spanx style="verb">?.co.uk</spanx> would both be problematic). Logs MUST NOT reject any precertificate that is overly redacted but which is otherwise considered compliant. It is expected that monitors will treat overly redacted precertificates as potentially misissued. TLS clients MAY reject a certificate whose corresponding precertificate would be overly redacted, perhaps using the same mechanism for determining whether a wildcard in a domain name of a certificate is too broad.
```
robstradling: "domains in the Public Suffix List".  The PSL isn't really a list of domains, per se.  publicsuffix.org says "The Public Suffix List is a list of all known public suffixes".  So how about...

`", such as <xref target="Public.Suffix.List">public suffixes</xref> (e.g. "

?

eranmes: Done

robstradling: Shall we also update the "overly redacted" language in this paragraph for logs, monitors and TLS clients?  If "overly redacted" is somewhat subjective for CAs, then it follows that it's somewhat subjective for logs, monitors and TLS clients too.  Something like this...
"Logs MUST NOT reject any precertificate that could be considered to be overly redacted. It is expected that monitors will treat precertificates that they consider to be overly redacted as potentially misissued. TLS clients MAY reject a certificate if they consider the corresponding precertificate to be overly redacted..."

robstradling: BTW, should "TLS clients MAY reject..." be changed to "TLS clients MAY treat...as non-compliant" ?

rfc6962-bis.xml
```diff
@@ -1643,13 +1643,7 @@ corrective action when a misissue is detected.
...
-certificate whose corresponding precertificate would be overly redacted, perhaps using the same mechanism for determining whether a wildcard in a domain name of a certificate is too broad.
+          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). CAs should take care not to redact domain name labels in precertificates too broadly, so that the entirety of the domain space below the unredacted part of the domain name may be controlled by multiple entities, such as <xref target="Public.Suffix.List">public suffixes</xref> (e.g. <spanx style="verb">?.com</spanx> and <spanx style="verb">?.co.uk</spanx> would both be problematic). Logs MUST NOT reject any precertificate that is overly redacted but which is otherwise considered compliant. It is expected that monitors will treat overly redacted precertificates as potentially misissued. TLS clients MAY reject a certificate whose corresponding precertificate would be overly redacted, perhaps using the same mechanism for determining whether a wildcard in a domain name of a certificate is too broad.
```
robstradling: I don't think "take care not to redact...*so* that the entirety of the domain space...may be controlled by multiple entities" expresses the intended meaning clearly.  I've had a go at rephrasing this section...

"Redaction of domain name labels carries the same risks as the use of wildcards (see Section 7.2 of &lt;xref target="RFC6125"/&gt;, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. &lt;spanx style="verb"&gt;?.com&lt;/spanx&gt;, &lt;spanx style="verb"&gt;?.co.uk&lt;/spanx&gt; and other &lt;xref target="Public.Suffix.List"&gt;public suffixes&lt;/xref&gt;), then the domain name may be considered by clients to be overly redacted.
</t>
<t>
CAs should take care to avoid overly redacting domain names in precertificates. It is expected that monitors will treat precertificates that contain overly redacted domain names as potentially misissued. TLS clients MAY reject a certificate if the &lt;xref target="reconstructing_tbscertificate"&gt;reconstructed TBSCertificate&lt;/xref&gt; contains any overly redacted domain names.whose corresponding precertificate would be overly redacted."

Note that I dropped the "perhaps using the same mechanism" clause.  This seems redundant now, given that we're citing the PSL as one way of doing it.

rfc6962-bis.xml
```diff
@@ -1643,13 +1643,7 @@ corrective action when a misissue is detected.
       </section>
       <section title="Redaction of Public Domain Name Labels">
```
robstradling: I suggest renaming this section "Overly Redacting Domain Name Labels" or "Avoiding Overly Redacting Domain Name Labels".

rfc6962-bis.xml
eranmes: Adopted your suggestion, with the difference of saying TLS clients MAY consider a certificate non-compliant if it's overly redacted. (they may do other things, of course, like rejecting it)

rfc6962-bis.xml

rfc6962-bis.xml
```diff
@@ -1641,16 +1641,12 @@ instead on interested parties, such as domain owners, to monitor them and take
...
-redacted precertificates as potentially misissued. TLS clients MAY reject a
-certificate whose corresponding precertificate would be overly redacted, perhaps using the same mechanism for determining whether a wildcard in a domain name of a certificate is too broad.
+          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. <spanx style="verb">?.com</spanx>, <spanx style="verb">?.co.uk</spanx> and other <xref target="Public.Suffix.List">public suffixes</xref>), then the domain name may be considered by clients to be overly redacted.
         </t>
+        <t>
+          CAs should take care to avoid overly redacting domain names in precertificates. It is expected that monitors will treat precertificates that contain overly redacted domain names as potentially misissued. TLS clients MAY consider a certificate as non-compliant if the <xref target="reconstructing_tbscertificate">reconstructed TBSCertificate</xref> contains any overly redacted domain names."</t>
```
eranmes: Nit (1) done.
Nit (2)  - I think this was a stray '"'

eranmes:
@robstradling PTAL and attach the LGTM label if you approve of the latest fix.

PR #132, opened 2016-03-09
Indicate logs SHOULD reuse SCTs.

eranmes:
To limit the log's ability of fingerprinting clients.


eranmes:
@benlaurie Added text per your suggestion offline.

benlaurie:
LGTM

PR #133, opened 2016-03-09
Fixing a few obsolete v1 references.

eranmes:


rfc6962-bis.xml
```diff
@@ -1103,7 +1103,7 @@ messages.
         <t>
           Note that this message is not signed -- the <spanx style="verb">entries</spanx> data can be
 verified by constructing the Merkle Tree Hash corresponding to a retrieved
-STH. All leaves MUST be v1 or v2. However, a compliant v1 client MUST NOT construe an
+STH. All leaves MUST be v1 or v2. However, a compliant v2 client MUST NOT construe an
```
robstradling: "All leaves MUST be v1 or v2".  Does that make sense?  Or do we need to require that (in order to conform to this version of the specification) all leaves MUST be v2?  (i.e. each log is either entirely v1 or entirely v2).

eranmes: You're right, it doesn't make a lot of sense - v2 logs should be entirely  v2.

eranmes: There are further problems with this sentence because LogEntryType doesn't exist anymore. Looking into re-writing it.

eranmes:
Done, PTAL.

rfc6962-bis.xml
```diff
@@ -1101,13 +1101,7 @@ messages.
...
         </t>
         <t>
-          Note that this message is not signed -- the <spanx style="verb">entries</spanx> data can be
-verified by constructing the Merkle Tree Hash corresponding to a retrieved
-STH. All leaves MUST be v1 or v2. However, a compliant v1 client MUST NOT construe an
-unrecognized LogEntryType value as an error. This means it
-may be unable to parse some entries, but note that each client can inspect the
-entries it does recognize as well as verify the integrity of the data by
-treating unrecognized leaves as opaque input to the tree.
+          Note that this message is not signed -- the <spanx style="verb">entries</spanx> data can be verified by constructing the Merkle Tree Hash corresponding to a retrieved STH. All leaves MUST be v2. However, a compliant v2 client MUST NOT construe an unrecognized TransItem value as an error. This means it may be unable to parse some entries, but note that each client can inspect the entries it does recognize as well as verify the integrity of the data by treating unrecognized leaves as opaque input to the tree.
```
robstradling: "unrecognized TransItem value".  It's specifically the type of the TransItem that may be unrecognized.

I suggest: s/TransItem value/TransItem type/

PR #134, opened 2016-03-14
Language change around SCTs with timestamp in the future.

eranmes:
Clarify that rejection of an SCT means treating it as non-valid.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1316,7 +1316,7 @@ but it is expected there will be a variety.
         </section>
         <section title="Validating SCTs">
           <t>
-            In addition to normal validation of the server certificate and its chain, TLS clients SHOULD validate each received SCT for which they have the corresponding log's metadata. To validate an SCT, a TLS client computes the signature input from the SCT data and the corresponding certificate, and then verifies the signature using the corresponding log's public key. TLS clients MUST reject SCTs whose timestamp is in the future.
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1316,7 +1316,7 @@ but it is expected there will be a variety.
         </section>
         <section title="Validating SCTs">
           <t>
-            In addition to normal validation of the server certificate and its chain, TLS clients SHOULD validate each received SCT for which they have the corresponding log's metadata. To validate an SCT, a TLS client computes the signature input from the SCT data and the corresponding certificate, and then verifies the signature using the corresponding log's public key. TLS clients MUST reject SCTs whose timestamp is in the future.
+            In addition to normal validation of the server certificate and its chain, TLS clients SHOULD validate each received SCT for which they have the corresponding log's metadata. To validate an SCT, a TLS client computes the signature input from the SCT data and the corresponding certificate, and then verifies the signature using the corresponding log's public key. TLS clients MUST NOT consider SCTs whose timestamp is in the future as valid.
```
robstradling: "consider...as" doesn't carry the desired meaning, I think.
http://ell.stackexchange.com/questions/11226/is-considered-always-followed-by-as

How about this?
"TLS clients MUST NOT consider valid any SCT whose timestamp is in the future."

https://www.google.co.uk/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=%22not+consider+valid+any%22

eranmes:
@robstradling PTAL.

PR #135, opened 2016-03-14
Ticket 153: Explain V1/V2 interoperability.

eranmes:
V1 and V2 SCTs can co-exist for X.509 certificates, but not precertificates.
I've briefly explored the option of allowing V2 SCTs to be valid for V1 precerts
by specifying that V2 clients should remove the X.509 extension with V1 SCTs
prior to checking V2 SCT signatures.

However it is unclear the complication is worth it as V1 clients will not
be able to use V2 SCTs, so the benefit is unclear.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -496,6 +496,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
     } TransType;
 
     struct {
+        Version version;
         TransType type;
         select (type) {
             case x509_entry: TimestampedCertificateEntryDataV2;
@@ -507,18 +508,11 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
             case consistency_proof: ConsistencyProofDataV2;
             case inclusion_proof: InclusionProofDataV2;
         } data;
-    } TransItemV2;
-
-    struct {
-        Version version;
-        select (version) {
-            case v1: TransItemV1;
-            case v2: TransItemV2;
-        }
-    } TransItem;</artwork>
+    } TransItem;
+          </artwork>
         </figure>
         <t>
-          <spanx style="verb">version</spanx> is the earliest version of this protocol to which the encapsulated data structure conforms. This document is v2. Note that <xref target="RFC6962">v1</xref> did not define <spanx style="verb">TransItem</spanx>, but this document specifies a mechanism (see <xref target="TransItemV1"/>) for v2 implementations to encapsulate existing v1 objects in the <spanx style="verb">TransItem</spanx> structure. Note also that, since each <spanx style="verb">TransItem</spanx> object is individually versioned, future revisions to this protocol could conceivably update some encapsulated data structures without having to update all of them.
+          <spanx style="verb">version</spanx> is the earliest version of this protocol to which the encapsulated data structure conforms. This document is v2. Note that <xref target="RFC6962">v1</xref> did not define <spanx style="verb">TransItem</spanx>, but this document provides guidelines (see <xref target="v1_coexistence"/>) on how v2 implementations can co-exist with v1 implementations. Note also that, since each <spanx style="verb">TransItem</spanx> object is individually versioned, the version should be increased only if changes to it are made that are not backwards-compatible. The addition of encapsulated data structures can be done by adding <spanx style="verb">TransType</spanx> values without increasing the version.
         </t>
         <t>
           <spanx style="verb">type</spanx> is the type of the encapsulated data structure. (Note that <spanx style="verb">TransType</spanx> combines the v1 type enumerations <spanx style="verb">LogEntryType</spanx>, <spanx style="verb">SignatureType</spanx> and <spanx style="verb">MerkleLeafType</spanx>). Future revisions of this protocol may add new <spanx style="verb">TransType</spanx> values.
@@ -1310,6 +1304,9 @@ but it is expected there will be a variety.
             </list>
             A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
           </t>
+          <t>
+            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_sct</spanx>), then the client SHOULD also remove embedded v1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded v1 and v2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
+          </t>
         </section>
         <section title="Validating SCTs">
           <t>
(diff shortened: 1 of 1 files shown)


eranmes:
@robstradling PTAL.

eranmes:
@benlaurie you too since it's a non-trivial change, you may have an opinion.

rfc6962-bis.xml
```diff
@@ -507,18 +508,11 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          <spanx style="verb">version</spanx> is the earliest version of this protocol to which the encapsulated data structure conforms. This document is v2. Note that <xref target="RFC6962">v1</xref> did not define <spanx style="verb">TransItem</spanx>, but this document provides guidelines (see <xref target="v1_coexistence"/>) on how v2 implementations can co-exist with v1 implementations. Note also that, since each <spanx style="verb">TransItem</spanx> object is individually versioned, future revisions to this protocol could conceivably update some encapsulated data structures without having to update all of them.
```
benlaurie: I realise this is pre-existing text, but I now find it confusing. If, for example, I added a new TransType, it would seem I would been to bump the version field ... however, that doesn't actually seem helpful.

eranmes: How about a more explicit statement saying the version should be changed if changes to the TransItem data structure are made that are not backwards compatible, while the addition of a type does not require a version bump?

benlaurie: That would work for me.

benlaurie:
LGTM apart from aforementioned confusing text.

benlaurie:
which can be addressed in another PR, btw.

rfc6962-bis.xml
```diff
@@ -1313,6 +1307,9 @@ but it is expected there will be a variety.
             </list>
             A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
           </t>
+          <t>
+            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_entry</spanx>), then the client SHOULD also remove embedded V1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded V1 and V2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
```
robstradling: Shouldn't that be: s/precert_entry/precert_sct/  ?

rfc6962-bis.xml
```diff
@@ -1856,71 +1853,35 @@ value="Montreal"/>
...
+    <section title="Supporting V1 and V2 simultaneously" anchor="v1_coexistence">
+      <t>
+        Certificate Transparency logs have to be either V1 (conforming to <xref target="RFC6962"/>) or V2 (conforming to this document), as the data structures are incompatible and so a V2 log could not issue a valid V1 SCT.
+      </t>
+      <t>
+        CT clients, however, can support V1 and V2 SCTs, for the same certificate, simultaneously, as V1 SCTs are delivered in different TLS, X.509 and OCSP extensions than V2 SCTs.
+      </t>
+      <t>
+        V1 and V2 SCTs for X.509 certificates can be validated independently.
+        For precertificates, V2 SCTs should be embedded in the TBSCertificate before submission of the TBSCertificate (inside a V1 precertificate, as described in Section 3.1. of <xref target="RFC6962"/>) so that TLS clients conforming to <xref target="RFC6962"/> but not this document are oblivious to the embedded V2 SCTs.
```
robstradling: "before submission" to where?  Presumably you mean "before submission...to a V1 log"?

rfc6962-bis.xml
```diff
@@ -1856,71 +1853,35 @@ value="Montreal"/>
...
-    } ConsistencyProofDataV1;
-
-    struct {
-        SHA256Hash log_id;
-        uint64 tree_size;
-        uint64 leaf_index;
-        SHA256Hash inclusion_path&lt;1..2^8-1&gt;;
-    } InclusionProofDataV1;</artwork>
-      </figure>
+    <section title="Supporting V1 and V2 simultaneously" anchor="v1_coexistence">
```
robstradling: Elsewhere in the document, except for the data structure names, we already use a lower-case "v" (i.e. "v1", "v2").  Let's be consistent.

rfc6962-bis.xml
```diff
@@ -1856,71 +1853,35 @@ value="Montreal"/>
...
+        An issuer can follow these steps to produce an X.509 certificate with valid V1 and V2 SCTs:
+        <list style="symbols">
+          <t>
+            Create a CMS precertificate as described in <xref target="Precertificates"/> and submit it to V2 logs.
+          </t>
+          <t>
+            Embed the obtained V2 SCTs in the TBSCertificate, as described in <xref target="cert_transinfo_extension"/>.
+          </t>
+          <t>
+            Create a V1 precertificate, as described in Section 3.1. of <xref target="RFC6962"/> and submit it to V1 logs.
```
robstradling: I don't think it's clear that "Create a V1 precertificate..." needs to use the TBSCertificate from the previous step.

How about:
"Use that TBSCertificate to create a v1 precertificate..."

rfc6962-bis.xml
```diff
@@ -1856,71 +1853,35 @@ value="Montreal"/>
...
+          </t>
+          <t>
+            Embed the V1 precertificates in the TBSCertificate, as described in Section 3.3. of <xref target="RFC6962"/>.
```
robstradling: Precertificates are not embedded in certificates!
SCTs (of the correct type) are embedded in certificates!

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: Good catch, thanks.

rfc6962-bis.xml
eranmes: Added 'to a v1 log'.

rfc6962-bis.xml

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: You are right, of course. Fixed.

eranmes:

rfc6962-bis.xml
```diff
@@ -512,7 +512,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          <spanx style="verb">version</spanx> is the earliest version of this protocol to which the encapsulated data structure conforms. This document is v2. Note that <xref target="RFC6962">v1</xref> did not define <spanx style="verb">TransItem</spanx>, but this document provides guidelines (see <xref target="v1_coexistence"/>) on how v2 implementations can co-exist with v1 implementations. Note also that, since each <spanx style="verb">TransItem</spanx> object is individually versioned, the version should be increased if changes to it are made that are not backwards-compatible. The addition of encapsulated data structures can be done by adding <spanx style="verb">TransType</spanx> values without increasing the version.
```
robstradling: This sounds unnecessarily complicated.  I'm growing increasingly convinced that we should conflate TransType and Version into a single TransTypeAndVersion enum (but let's do that in a separate ticket / PR).

rfc6962-bis.xml
```diff
@@ -1853,32 +1853,32 @@ value="Montreal"/>
...
       </t>
       <t>
-        V1 and V2 SCTs for X.509 certificates can be validated independently.
-        For precertificates, V2 SCTs should be embedded in the TBSCertificate before submission of the TBSCertificate (inside a V1 precertificate, as described in Section 3.1. of <xref target="RFC6962"/>) so that TLS clients conforming to <xref target="RFC6962"/> but not this document are oblivious to the embedded V2 SCTs.
-        An issuer can follow these steps to produce an X.509 certificate with valid V1 and V2 SCTs:
+        v1 and v2 SCTs for X.509 certificates can be validated independently.
+        For precertificates, v2 SCTs should be embedded in the TBSCertificate before submission of the TBSCertificate (inside a v1 precertificate, as described in Section 3.1. of <xref target="RFC6962"/>) to a v1 log so that TLS clients conforming to <xref target="RFC6962"/> but not this document are oblivious to the embedded v2 SCTs.
+        An issuer can follow these steps to produce an X.509 certificate with valid v1 and v2 SCTs:
```
robstradling: Nit: s/valid/embedded/

Valid is implied, I think.  Not all SCTs are embedded in certs, so this should be explicitly specified.

rfc6962-bis.xml
eranmes: SGTM - would you like to track this in a ticket?

rfc6962-bis.xml

eranmes:
@robstradling PTAL.

rfc6962-bis.xml
robstradling: Done.  https://trac.tools.ietf.org/wg/trans/trac/ticket/157

rfc6962-bis.xml
```diff
@@ -507,18 +508,11 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          <spanx style="verb">version</spanx> is the earliest version of this protocol to which the encapsulated data structure conforms. This document is v2. Note that <xref target="RFC6962">v1</xref> did not define <spanx style="verb">TransItem</spanx>, but this document provides guidelines (see <xref target="v1_coexistence"/>) on how v2 implementations can co-exist with v1 implementations. Note also that, since each <spanx style="verb">TransItem</spanx> object is individually versioned, the version should be increased if changes to it are made that are not backwards-compatible. The addition of encapsulated data structures can be done by adding <spanx style="verb">TransType</spanx> values without increasing the version.
```
benlaurie: s/increased if/increased only if/

PR #139, opened 2016-03-17
Ticket 155: Describe how to match SCTs to certificates.

robstradling:
Suggest evaluating the certificates in the order they were sent by the TLS server.
Clarify that TLS clients MUST NOT consider invalid any SCT or inclusion proof until each potentially corresponding certificate has been evaluated.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1319,12 +1319,15 @@ but it is expected there will be a variety.
             In addition to normal validation of the server certificate and its chain, TLS clients SHOULD validate each received SCT for which they have the corresponding log's metadata. To validate an SCT, a TLS client computes the signature input from the SCT data and the corresponding certificate, and then verifies the signature using the corresponding log's public key. TLS clients MUST NOT consider valid any SCT whose timestamp is in the future.
           </t>
           <t>
-            Since each received SCT may be valid either for the server certificate itself or for a <xref target="name_constrained">suitable name-constrained intermediate</xref>, TLS clients MAY need to use trial and error to determine which certificate in the chain is the certificate to which a received SCT corresponds.
+            Before considering any SCT to be invalid, the TLS client MUST attempt to validate it against the server certificate and against each of the zero or more <xref target="name_constrained">suitable name-constrained intermediates</xref> in the chain. These certificates may be evaluated in the order they appear in the chain, or, indeed, in any order.
           </t>
         </section>
         <section title="Validating inclusion proofs">
           <t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1319,12 +1319,15 @@ but it is expected there will be a variety.
             In addition to normal validation of the server certificate and its chain, TLS clients SHOULD validate each received SCT for which they have the corresponding log's metadata. To validate an SCT, a TLS client computes the signature input from the SCT data and the corresponding certificate, and then verifies the signature using the corresponding log's public key. TLS clients MUST NOT consider valid any SCT whose timestamp is in the future.
           </t>
           <t>
-            Since each received SCT may be valid either for the server certificate itself or for a <xref target="name_constrained">suitable name-constrained intermediate</xref>, TLS clients MAY need to use trial and error to determine which certificate in the chain is the certificate to which a received SCT corresponds.
+            Before considering any SCT to be invalid, the TLS client MUST attempt to validate it against the server certificate and against each of the zero or more <xref target="name_constrained">suitable name-constrained intermediates</xref> in the chain. These certificates may be evaluated in the order they were sent by the TLS server, or, indeed, in any order.
```
eranmes: I think the final sentence ("These certificates may be evaluated in the order they were sent by the TLS server, or, indeed, in any order.") should be removed:
(1) it may not be feasible to require that TLS clients evaluate SCTs on the chain received, rather than the chain built by the client, as different clients have different heuristics for building chains.
(2) What if a server sends a name-constrained intermediate that is not even a part of the chain, but sends a valid SCT for it? I think that in that case the client should not consider the leaf to have a valid SCT covering its chain.

To sum it up, I suggest we not get involved with the way clients build their chains and simply say that the client should attempt validating the SCT against each of the suitable name-constrained intermediates in it.

eranmes:
LGTM except the one comment I've left .

rfc6962-bis.xml
robstradling: This PR needs to provide "a description of one way a client could match SCTs to a certificate in the chain", because "use trial and error" was considered unsuitably vague.  (That's why I wrote the "in the order they were sent by the TLS server" suggestion).

Steve Kent was looking for a suggested algorithm such as "Start with the server certificate, then look at its issuer, then look at its issuer, and so on, until you reach the root", wasn't he?

Just saying "that the client should attempt validating the SCT against each of the suitable name-constrained intermediates in it" describes every conceivable way.  It's not an example of "one way".

robstradling: How about:
"These certificates may be evaluated in the order they appear in the chain, or, indeed, in any order"
?

eranmes: Sounds good to me.

eranmes:
LGTM with the proposed change.

PR #141, opened 2016-03-18
Ticket 141: Further expand auditing

eranmes:
Incorporate some of the comments in ticket 141, describing what aspects of log
behaviour should be audited.
Note that while some are described in the auditing section, others are
in the security considerations section.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1418,6 +1418,23 @@ but it is expected there will be a variety.
           Auditing ensures that the current published state of a log is reachable from previously published states that are known to be good, and that the promises made by the log in the form of SCTs have been kept.  Audits are performed by monitors or TLS clients.
         </t>
         <t>
+          In particular, there are four log behaviour properties that should be checked:
+          <list style="symbols">
+            <t>
+              The Maximum Merge Delay (MMD).
+            </t>
+            <t>
+              The STH Frequency Count.
+            </t>
+            <t>
+              The append-only property.
+            </t>
+            <t>
+              The consistency of the log view presented to all query sources.
+            </t>
+          </list>
+        </t>
+        <t>
           A benign, conformant log publishes a series of STHs over time, each derived from the previous STH and the submitted entries incorporated into the log since publication of the previous STH.  This can be proven through auditing of STHs.  SCTs returned to TLS clients can be audited by verifying against the accompanying certificate, and using Merkle Inclusion Proofs, against the log's Merkle tree.
         </t>
         <t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1656,13 +1674,12 @@ this can be done.
...
+the proofs from the log) or communicate with the log via proxies.
         </t>
         <t>
-          Violation of the append-only property can be detected by clients comparing their instances of the Signed Tree Heads. As soon as two conflicting Signed Tree Heads for the same log are detected, this is cryptographic proof of that log's misbehavior. There are various ways this could be done, for example via gossip (see <xref target="I-D.ietf-trans-gossip"/>) or peer-to-peer communications or by sending STHs to monitors (who could then directly check against their own copy of the relevant log).
+          Violation of the append-only property or the STH issuance rate limit can be detected by clients comparing their instances of the Signed Tree Heads. As soon as two conflicting Signed Tree Heads for the same log are detected, this is cryptographic proof of that log's misbehavior. There are various ways this could be done, for example via gossip (see <xref target="I-D.ietf-trans-gossip"/>) or peer-to-peer communications or by sending STHs to monitors (who could then directly check against their own copy of the relevant log).
```
benlaurie: This is a little bit of an oversimplifcation. The proof of misbehaviour is that one of the STHs doesn't match a calculated STH from a copy of the log. You never get a proof of inconsistency between two STHs, you just can't get a proof of consistency, so a log, when asked for one, will presumably not respond, or respond with an incorrect proof (which still doesn't prove anything).

eranmes: I've re-wrote this sentence (and moved it to the end of the paragraph) to be more accurate, per your comment.

eranmes:

rfc6962-bis.xml
```diff
@@ -1677,7 +1677,7 @@ can request the proof from a trusted auditor (since anyone can compute
...
         </t>
         <t>
-          Violation of the append-only property or the STH issuance rate limit can be detected by clients comparing their instances of the Signed Tree Heads. As soon as two conflicting Signed Tree Heads for the same log are detected, this is cryptographic proof of that log's misbehavior. There are various ways this could be done, for example via gossip (see <xref target="I-D.ietf-trans-gossip"/>) or peer-to-peer communications or by sending STHs to monitors (who could then directly check against their own copy of the relevant log).
+          Violation of the append-only property or the STH issuance rate limit can be detected by clients comparing their instances of the Signed Tree Heads. There are various ways this could be done, for example via gossip (see <xref target="I-D.ietf-trans-gossip"/>) or peer-to-peer communications or by sending STHs to monitors (who could then directly check against their own copy of the relevant log). A proof of misbehavior in such cases would be a series of STHs that were issued too closely together or an STH with a root hash that does not match the one calculated from a copy of the log.
```
benlaurie: This still mentions append-only, but only talks about STH rate.

benlaurie: Oh it does. Sorry! Slightly weird to be conflating the two.

eranmes: Shall I separate them into separate sentences?

rfc6962-bis.xml
```diff
@@ -1654,13 +1672,12 @@ this can be done.
...
+          Violation of the append-only property or the STH issuance rate limit can be detected by clients comparing their instances of the Signed Tree Heads. There are various ways this could be done, for example via gossip (see <xref target="I-D.ietf-trans-gossip"/>) or peer-to-peer communications or by sending STHs to monitors (who could then directly check against their own copy of the relevant log). A proof of misbehavior in such cases would be a series of STHs that were issued too closely together, proving violation of the STH issuance rate limit, or an STH with a root hash that does not match the one calculated from a copy of the log, proving entries in the logs were changed since that STH was issued.
```
benlaurie: It doesn't prove that the entries were changed _since_ the STH was issued (i.e. it could've been before it was issued, but after the copy was taken). So, I would just say "proving violation of the append-only property".

eranmes:
@benlaurie PTAL.

PR #142, opened 2016-03-18
Ticket 156: Requiring SCTs in addition to inclusion proofs

eranmes:
…proofs.

The gist of the reasoning is that without the timestamp and extensions fields
from the SCT, it is not possible to calculate the leaf hash that would
be used together with the rest of the nodes from the inclusion proof to
calculate the root hash.

To make implementation of clients simpler (i.e. clients would only need
to check signatures from SCTs rather than having to check them over STHs
as well and implement inclusion checking) this PR suggests mandating
presence of SCTs, while pointing out the benefit of supplying both
SCT and inclusion proof.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -293,7 +293,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         </section>
         <section title="Signatures" anchor="signatures">
           <t>
-            Various data structures are signed. A log MUST use one of the signature algorithms defined in the <xref target="signature_algorithms"/> section.
+            Various data structures are signed. A log MUST use one of the signature algorithms defined in the <xref target="signature_algorithms"/>.
           </t>
         </section>
       </section>
@@ -1145,10 +1145,13 @@ entry specified by <spanx style="verb">start</spanx>.
     </section>
     <section title="TLS Servers" anchor="tls_servers">
       <t>
-        TLS servers MUST use at least one of the three mechanisms listed below to present one or more SCTs or inclusion proofs from one or more logs to each TLS client during TLS handshakes, where each SCT or inclusion proof corresponds to the server certificate or to a name-constrained intermediate the server certificate chains to. Three mechanisms are provided because they have different tradeoffs.
+        TLS servers MUST use at least one of the three mechanisms listed below to present one or more SCTs from one or more logs to each TLS client during TLS handshakes, where each SCT corresponds to the server certificate or to a name-constrained intermediate the server certificate chains to. TLS servers SHOULD present corresponding inclusion proofs together with Signed Tree Heads (<spanx style="verb">TransItem</spanx> of types <spanx style="verb">inclusion_proof_v2</spanx> or <spanx style="verb">signed_tree_head_v2</spanx>) in addition to the SCTs. When presented, the inclusion proof MUST be for computing the root in the accompanying Signed Tree Head (presenting inclusion proofs increases client privacy by reducing the inclusion checking to STH consistency checking and reduces load on log servers).
+      </t>
+      <t>
+        Three mechanisms are provided because they have different tradeoffs.
         <list style="symbols">
           <t>
-            A TLS extension (Section 7.4.1.4 of <xref target='RFC5246'/>) with type <spanx style="verb">transparency_info</spanx> (see <xref target="tls_transinfo_extension"/>). This mechanism allows TLS servers to participate in CT without the cooperation of CAs, unlike the other two mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
+            A TLS extension (Section 7.4.1.4 of <xref target='RFC5246'/>) with type <spanx style="verb">transparency_info</spanx> (see <xref target="tls_transinfo_extension"/>). This mechanism allows TLS servers to participate in CT without the cooperation of CAs, unlike the other two mechanisms. It also allows SCTs to be updated on the fly.
           </t>
           <t>
             An <xref target='RFC6960'>Online Certificate Status Protocol (OCSP)</xref> response extension (see <xref target="ocsp_transinfo_extension"/>), where the OCSP response is provided in the <spanx style="verb">CertificateStatus</spanx> message, provided that the TLS client included the <spanx style="verb">status_request</spanx> extension in the (extended) <spanx style="verb">ClientHello</spanx> (Section 8 of <xref target='RFC6066'/>). This mechanism, popularly known as OCSP stapling, is already widely (but not universally) implemented. It also allows SCTs and inclusion proofs to be updated on the fly.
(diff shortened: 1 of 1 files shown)


eranmes:
To be clear, that is just one proposal to raise the problem I described on the ticket. Not locked into solving it this way.

rfc6962-bis.xml
```diff
@@ -1321,13 +1324,13 @@ but it is expected there will be a variety.
...
           </t>
           <t>
-            Before considering any inclusion proof to be invalid, the TLS client MUST attempt to validate it against the server certificate and against each of the zero or more <xref target="name_constrained">suitable name-constrained intermediates</xref> in the chain. These certificates may be evaluated in the order they appear in the chain, or, indeed, in any order.
+            Alternatively, in case the TLS client has received an inclusion proof (and an STH) alongside the SCT, it can proceed to verifying the inclusion proof to the provided STH. The client then has to verify consistency between the provided STH and an STH it knows about, which is less sensitive from a privacy perspective.
```
benlaurie: s/in case/if/

rfc6962-bis.xml
```diff
@@ -1153,10 +1153,13 @@ entry specified by <spanx style="verb">start</spanx>.
     </section>
     <section title="TLS Servers" anchor="tls_servers">
       <t>
-        TLS servers MUST use at least one of the three mechanisms listed below to present one or more SCTs or inclusion proofs from one or more logs to each TLS client during TLS handshakes, where each SCT or inclusion proof corresponds to the server certificate or to a name-constrained intermediate the server certificate chains to. Three mechanisms are provided because they have different tradeoffs.
+        TLS servers MUST use at least one of the three mechanisms listed below to present one or more SCTs from one or more logs to each TLS client during TLS handshakes, where each SCT corresponds to the server certificate or to a name-constrained intermediate the server certificate chains to. TLS servers MAY present relevant inclusion proofs together with Signed Tree Heads(<spanx style="verb">TransItem</spanx> of types <spanx style="verb">inclusion_proof</spanx> or <spanx style="verb">signed_tree_head</spanx>) in addition to the SCTs. When presented, the inclusion proof SHOULD be for computing the root in the accompanying Signed Tree Head.
```
benlaurie: Let's make it SHOULD, not MAY. And also say that it SHOULD use a recent STH. And that the STH MUST correspond to the inclusion proof.

And maybe say why a server would do this (to increase client privacy and reduce log load).

robstradling: Ben, what's your view on http://trac.tools.ietf.org/wg/trans/trac/ticket/156#comment:4 ?

benlaurie: Yes, that's a good idea ... reducing complexity in the client always seems like a good plan.

Eran wants clients to be able to only check the SCT, but that would still be possible with a dedicated data structure.

robstradling: "to be able to only check the SCT" wasn't what I meant.  I'd also like it to be possible for servers to only send, and for clients to only check, inclusion proofs too.  Because why would a client need to verify the promise of inclusion if it verifies the actual inclusion?
Sending multiple SCT signatures in the handshake that the client doesn't need to check seems rather wasteful.

eranmes:
@robstradling IMHO even by checking an embedded inclusion proof, it'd still be a "promise of inclusion" until the client can reconcile the STH in the inclusion proof with an STH it knows about.

Regardless, I would not mind a dedicated data structure, as suggested in http://trac.tools.ietf.org/wg/trans/trac/ticket/156#comment:4.
What I feel strongly about is for that data structure to contain the signature from the SignedCertificateTimestampDataV2 structure in that dedicated data structure (in addition to the timestamp and sct_extensions which are essential, as I've detailed in the ticket).

The reason being the amount of added code in clients (TLS clients as well as CA software) needed to verify inclusion proofs is significantly more (and more complex) than verifying SCT signature over the serialized data.
Given the complexity we've added to clients for dealing with name-redaction, it'd mean quite a lot of code just to check if a certificate is CT compliant or not.
That's why I'd like clients not to have to implement inclusion checking just for checking CT compliance and would like the SCT signature to be present.

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: Addressed Ben's original comment.

eranmes:
@robstradling  and @benlaurie  PTAL.

robstradling:
I'm going to do a U-turn on this, but for a different reason than client complexity.

If a client does check inclusion proofs and finds one to be _invalid_, the client should also check the corresponding SCT, because if that SCT is _valid_, that's evidence that the log has _not_ kept its promise.

rfc6962-bis.xml
```diff
@@ -1147,10 +1147,13 @@ entry specified by <spanx style="verb">start</spanx>.
...
+        TLS servers MUST use at least one of the three mechanisms listed below to present one or more SCTs from one or more logs to each TLS client during TLS handshakes, where each SCT corresponds to the server certificate or to a name-constrained intermediate the server certificate chains to. TLS servers SHOULD present relevant inclusion proofs together with Signed Tree Heads(<spanx style="verb">TransItem</spanx> of types <spanx style="verb">inclusion_proof</spanx> or <spanx style="verb">signed_tree_head</spanx>) in addition to the SCTs. When presented, the inclusion proof MUST be for computing the root in the accompanying Signed Tree Head (presenting inclusion proofs increases client privacy by reducing the inclusion checking to STH consistency checking and reduces load on log servers).
```
robstradling: I suggest: s/relevant/corresponding/

Nit: Need space after "Heads".

Nit: "inclusion_proof" and "signed_tree_head" now have a "_v2" suffix.  (I'm calling this a nit 'cos this text will need to change again anyway once we define that new data structure that will hold SCT + inclusion proof + STH - but let's do that in a separate ticket and separate PR).

rfc6962-bis.xml
```diff
@@ -1319,13 +1322,13 @@ but it is expected there will be a variety.
...
+            Alternatively, if the TLS client has received an inclusion proof (and an STH) alongside the SCT, it can proceed to verifying the inclusion proof to the provided STH. The client then has to verify consistency between the provided STH and an STH it knows about, which is less sensitive from a privacy perspective.
           </t>
           <t>
-            After validating a received SCT, a TLS client MAY request a corresponding inclusion proof (if one is not already available) and then verify it. An inclusion proof can be requested directly from a log using <spanx style="verb">get-proof-by-hash</spanx> (<xref target="get-proof-by-hash"/>) or <spanx style="verb">get-all-by-hash</spanx> (<xref target="get-all-by-hash"/>), but note that this will disclose to the log which TLS server the client has been communicating with.
+            TLS clients should also verify each received inclusion proof (see <xref target="verify_inclusion"/>) for which they have the corresponding log's metadata, to audit the log and gain confidence that the certificate is logged.
```
robstradling: That "should" was previously a "SHOULD".  Why the change?

rfc6962-bis.xml

rfc6962-bis.xml

eranmes:
@robstradling PTAL. Once that PR is in I'll re-assign ticket 156 to you so you can add the structure you described.

PR #143, opened 2016-03-21
Ticket 157: Conflate Version and TransType

robstradling:


rfc6962-bis.xml
```diff
@@ -486,45 +486,43 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          Note that <spanx style="verb">VersionedTransType</spanx> combines the <xref target="RFC6962">v1</xref> type enumerations <spanx style="verb">Version</spanx>, <spanx style="verb">LogEntryType</spanx>, <spanx style="verb">SignatureType</spanx> and <spanx style="verb">MerkleLeafType</spanx>. Note also that v1 did not define <spanx style="verb">TransItem</spanx>, but this document provides guidelines (see <xref target="v1_coexistence"/>) on how v2 implementations can co-exist with v1 implementations.
         </t>
         <t>
-          <spanx style="verb">data</spanx> is the encapsulated data structure. The various structures named with the <spanx style="verb">DataV2</spanx> suffix are defined in later sections of this document.
+          Future versions of this protocol may add new <spanx style="verb">VersionedTransType</spanx> values for new data structures, but need only do this for existing data structures that are changed in a manner that is backwards-incompatible.
```
eranmes: A bit confused - you're saying here that new VersionedTransTypes may be added for new data structures, ... but only do this for existing data structures that are changed?
Are you saying that new VersionedTransTypes values may be added only if we want to change one of the data structures defined in this document in a non backwards-compatible way but new values may not be added for new data structures that do not serve the same purpose as one of the predefined ones?

eranmes:
LGTM with one minor comment - once addressed, can be merged.

rfc6962-bis.xml
robstradling: "new data structures" in my text is intended to mean (as you put it) "new data structures that do not serve the same purpose as one of the predefined ones".

"need only do this for existing data structures" is meant to be interpreted as "need only add new VersionTransType values for existing data structures".  Is that not clear?

robstradling:
@eranmes PTAL

eranmes:
Still LGTM.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/919489291b7c9556d70e91e830045d54f7441f1f

PR #144, opened 2016-03-25
Ticket 156: New data structure for SCT + inclusion proof + STH.

robstradling:


eranmes:
LGTM. @benlaurie please take a look - if that's OK with you I will merge this PR after fixing the conflicts (which is just replacing the TLS servers section with Rob's section from this PR after adding 'full' to the TLS handshake description).

rfc6962-bis.xml
```diff
@@ -1195,7 +1203,34 @@ entry specified by <spanx style="verb">start</spanx>.
...
+        <t>
+          <spanx style="verb">sth</spanx> is the encapsulated data structure from an STH that was signed by the same log as <spanx style="verb">sct</spanx>.
+        </t>
+        <t>
+          <spanx style="verb">inclusion_proof</spanx> is the encapsulated data structure from an inclusion proof that corresponds to <spanx style="verb">sct</spanx> and can be used to compute the root in <spanx style="verb">sth</spanx>.
+        </t>
+      </section>
+      <section title="Presenting SCTs only">
+        <t>
+          Presenting inclusion proofs and STHs in the TLS handshake helps to protect the client's privacy (see <xref target="validating_inclusion_proofs"/>) and reduces load on log servers. However, if a TLS server is unable to obtain an inclusion proof and STH that correspond to an SCT, then it MAY include <spanx style="verb">TransItem</spanx> structures of type <spanx style="verb">x509_sct_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> in the <spanx style="verb">TransItemList</spanx>.
```
benlaurie: This should be a MUST, surely?

rfc6962-bis.xml
```diff
@@ -1195,7 +1203,34 @@ entry specified by <spanx style="verb">start</spanx>.
...
+        </t>
+      </section>
+      <section title="transparency_info TLS Extension" anchor="tls_transinfo_extension">
         <t>
           Provided that a TLS client includes the <spanx style="verb">transparency_info</spanx> extension type in the ClientHello, the TLS server MAY include the <spanx style="verb">transparency_info</spanx> extension in the ServerHello with <spanx style="verb">extension_data</spanx> set to a <spanx style="verb">TransItemList</spanx>. The TLS server SHOULD ignore any <spanx style="verb">extension_data</spanx> sent by the TLS client. Additionally, the TLS server MUST NOT process or include this extension when a TLS session is resumed, since session resumption uses the original session information.
```
benlaurie: MAY should be SHOULD?

rfc6962-bis.xml
eranmes: Yes, done.

rfc6962-bis.xml
eranmes: I can see a few reasons why a (CT-aware) TLS server would chose not to do this:
- It has no SCTs for a particular certificate it's presenting.
- It knows SCTs are already embedded in the certificate (or provided in the stapled OCSP response) so no extra SCTs are needed.

Under these circumstances do you still think we should switch this to SHOULD? Is your intention to require TLS servers to send a TransItemList of size 0 rather than send nothing at all?

eranmes: To capture an offline discussion: Since there's a MUST earlier the MAY would be changed to a SHOULD (and it fits with the paragraph's context).

eranmes:
Closing this PR in favour of PR #148 .

PR #147, opened 2016-03-25
Fix formatting.

robstradling:
Try running "xml2rfc --html" without this fix.  The "Reconstructing the TBSCertificate" section looks very odd!


eranmes:
Yeah, noticed that too but wasn't sure how to fix.

PR #148, opened 2016-03-29
Re-submission of PR 144: Ticket 156: New data structure for SCT + inclusion proof + STH.

eranmes:
So I can fix Ben's comments in Rob's absence.


eranmes:
@benlaurie please take another look, addressed all your comments from PR #144 .

PR #150, opened 2016-05-12
Remove reference targets which are never referenced.

ln5:


ln5:
Ah. Doing this with my NORDUnet hat on.

pphaneuf:
You might need to change the author of the commit, with some `git commit --amend` tomfoolery?

ln5:
I doubt any email address controlled by me is in Google's CLA database.
I'm letting you know the company's name as per instructions from the bot: NORDUnet A/S.

pphaneuf:
Hmm, we don't appear to have NORDUnet in the corporate CLA signers? I'll have to look into this...

I believe that for corporate CLAs, there's some way for them to list the committer emails that they cover, so the CLA bot would recognise you? But since I don't even see NORDUnet in there, that's not a good start! :wink:

pphaneuf:
Hmm, I can't find NORDUnet either in the current system, or in the old system (before the bot)... I think you'd need someone from NORDUnet to do it? It should be here, for both cases (this should also tell you if you're covered or not, and let you add email aliases and such, as well, once you are indeed covered):

https://cla.developers.google.com/

pphaneuf:
It does look like you used to have an "@nordu.net" email address, though (from a commit you've submitted to the code repo)? But doesn't change that I can't find NORDUnet in either the old or the new corporate CLA databases, sorry...

eranmes:
@ln5 can you confirm if you think NORDUnet has signed the CLA or intends to do so?
(Not sure it's needed, checking internally)

ln5:
@eranmes NORDUnet signed a "software grant and corporate contributer license agreement v1.1" on 2014-06-02. I think it was sent to Google.

eranmes:
Thanks, I'll follow up.

richsalz:
@robstradling you can close this, it's obsolete.

PR #151, opened 2016-05-18
Add get sct entry, DNS I-D fixes

benlaurie:
Sorry, two PRs in one, really, but both small...

@eranmes @robstradling

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -12,7 +12,7 @@
 <?rfc subcompact="no"?>
 <?rfc sortrefs="yes" ?>
 
-<rfc ipr="trust200902" category="std" submissionType="IETF" docName="draft-ietf-trans-rfc6962-bis-13">
+<rfc ipr="trust200902" category="std" submissionType="IETF" docName="draft-ietf-trans-rfc6962-bis-15">
 
   <front>
     <title>Certificate Transparency</title>
@@ -92,7 +92,7 @@ public certification authorities (CAs).
         Each log consists of certificate chains, which can be submitted by anyone. It is expected that public CAs will contribute all their newly issued certificates to one or more logs, however certificate holders can also contribute their own certificate chains, as can third parties. In order to avoid logs being rendered useless by the submission of large numbers of spurious certificates, it is required that each chain ends with a trust anchor that is accepted by the log. When a chain is accepted by a log, a signed timestamp is returned, which can later be used to provide evidence to TLS clients that the chain has been submitted. TLS clients can thus require that all certificates they accept as valid are accompanied by signed timestamps.
       </t>
       <t>
-        Those who are concerned about misissue can monitor the logs, asking
+        Those who are concerned about misissuance can monitor the logs, asking
 them regularly for all new entries, and can thus check whether domains they are
 responsible for have had certificates issued that they did not expect. What
 they do with this information, particularly when they find that a misissuance
@@ -293,7 +293,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         </section>
         <section title="Signatures" anchor="signatures">
           <t>
-            Various data structures are signed. A log MUST use one of the signature algorithms defined in the <xref target="signature_algorithms"/> section.
+            Various data structures are signed. A log MUST use one of the signature algorithms defined in the <xref target="signature_algorithms"/>.
           </t>
         </section>
       </section>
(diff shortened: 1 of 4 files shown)


rfc6962-bis.xml
```diff
@@ -1130,6 +1130,43 @@ entry specified by <spanx style="verb">start</spanx>.
...
+              <list style="hanging">
+                <t hangText="entry:">
+		  0-based index of the log entry corresponding to the supplied SCT.
+                </t>
+              </list>
+            </t>
+            <t hangText="Error codes:">
+              <list style="hanging">
+                <t hangText="bad signature">
+                  <spanx style="verb">sct</spanx> is not signed by this log.
```
robstradling: s/this log/this log's private key/

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,43 @@ entry specified by <spanx style="verb">start</spanx>.
...
+      <section title="Get Entry Number for SCT">
+	<t>
+	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-sct
+	</t>
+	<t>
+          <list style="hanging">
+            <t hangText="Inputs:">
+              <list style="hanging">
+                <t hangText="sct:">
+                  A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">x509_sct_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> allegedly from this log.
```
robstradling: s/from this log/signed by this log's private key/

"from this log" is too permissive.  A cert containing embedded SCTs from Log A and Log B might be logged to Log C.  If a client calls get-entries to fetch the corresponding log entry from Log C, then it will have obtained SCTs "from this log" (Log C) signed by the private keys of Logs A and B.

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,43 @@ entry specified by <spanx style="verb">start</spanx>.
...
+		</t>
+                <t hangText="not found">
+                  <spanx style="verb">sct</spanx> does not correspond to an entry that is currently available.
+		</t>
+	      </list>
+	    </t>
+          </list>
+        </t>
+	<t>
+	  Note that any SCT with a valid signature MUST have a corresponding entry in the log, but it may not be retrievable until the MMD has passed since the SCT was issued.
```
robstradling: A valid signature produced using which private key?

How about: s/with a valid signature/signed by the log's private key/

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,43 @@ entry specified by <spanx style="verb">start</spanx>.
         <t>See <xref target="verify_hash" /> for an outline of how to use a complete list of <spanx style="verb">leaf_input</spanx> entries to verify the <spanx style="verb">root_hash</spanx>.</t>
       </section>
 
+      <section title="Get Entry Number for SCT">
+	<t>
+	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-sct
```
eranmes: As discussed offline, let's make this an optional API. 
My argument for not making it mandatory is that while it would ease lookup in some cases, it is not an essential part of the protocol (i.e. the same functionality can be obtained by downloading the entire log) and adds a burden on log implementation.

eranmes:
DNS I-D changes LGTM, if you'd like to merge just those.

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,86 @@ entry specified by <spanx style="verb">start</spanx>.
         <t>See <xref target="verify_hash" /> for an outline of how to use a complete list of <spanx style="verb">leaf_input</spanx> entries to verify the <spanx style="verb">root_hash</spanx>.</t>
       </section>
 
+      <section title="Get Entry Number for SCT">
```
eranmes: Can we have (Optional) in the title?

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,86 @@ entry specified by <spanx style="verb">start</spanx>.
...
+	<t>
+	  Logs MAY implement this message. It is not required for correct operation of logs or their clients, but may be convenient in some circumstances, since an SCT cannot be validated with also knowing which chain it applies to. Note that mirrors can implement this even if the log they are mirroring does not.
```
eranmes: s/with also knowing/without also knowing/

rfc6962-bis.xml
eranmes: Is it necessary for the MUST here to be imperative? there's no way to implement a correct log without it, so I'd settle for a 'must' here.

rfc6962-bis.xml
```diff
@@ -1130,6 +1130,86 @@ entry specified by <spanx style="verb">start</spanx>.
...
+	</t>
+	<t>
+	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-certificate
+	</t>
+	<t>
+          <list style="hanging">
+            <t hangText="Inputs:">
+              <list style="hanging">
+                <t hangText="hash:">
+                  A base64 encoded HASH of a <spanx style="verb">TBSCertificate</spanx>.
```
eranmes: Since the log entry both for X.509 certs and precerts now has the issuer_key_hash in addition to the TBSCertificate, does it make sense to require it here as well?
May not be, just bringing it up for discussion.

rfc6962-bis.xml
benlaurie: We use "signed by this log" throughout the I-D.

rfc6962-bis.xml
benlaurie: See above. If this really is a problem, we should check all the other occurrences.

robstradling: OK, I can live with "signed by this log".

But please change "from this log" to "signed by this log".

rfc6962-bis.xml
benlaurie: Hmm ... I don't mind either way, but the MUST really is a MUST, as in it MUST, even if that happens to be demonstrable from other MUSTs.

rfc6962-bis.xml
robstradling: OK.  Leave this as is then.

rfc6962-bis.xml

rfc6962-bis.xml
benlaurie: @robstradling I went with "signed by a log".

rfc6962-bis.xml
benlaurie: I was thinking about that, will move to a new section.

rfc6962-bis.xml
benlaurie: No, because that requires the client to have more than the least required to retrieve the entry, which seems pointless.

rfc6962-bis.xml
robstradling: Could you make the next part of the sentence say "...corresponding entry in **that** log" then?

ISTM that "corresponding entry in the log" could be taken to mean any and every conceivable log.

rfc6962-bis.xml
eranmes: Is it worth explicitly calling out that certificates that were logged with redacted DNS labels should be looked up in their redacted form?

eranmes:
LGTM, you'll need to rebase though.

rfc6962-bis.xml
benlaurie: Hmm, that seems like a good idea.

benlaurie:
Merged at cf64f46.

PR #154, opened 2016-05-20
There should be exactly one SCT for each entry.

eranmes:
Attempting to clarify that, because the timestamp in the SCT is a part of the leaf data (it's in TimestampedCertificateEntryDataV2) and, because we use deterministic signatures, different SCTs for the same submission (which imply different timestamps) must be incorporated into the tree independently and there should be a log entry for each.
It is not enough to only incorporate one submission + SCT, since clients that will observe other SCTs will reconstruct a different TimestampedCertificateEntryDataV2 than the one in the log, and so will have a leaf hash for which the log can't prove inclusion.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -416,7 +416,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         A log is a single, append-only Merkle Tree of submitted certificate and precertificate entries.
       </t>
       <t>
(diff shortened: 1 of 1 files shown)


eranmes:
Also @benlaurie .

rfc6962-bis.xml
```diff
@@ -1091,7 +1091,7 @@ messages.
                       The base64 encoded log entry (see <xref target="log_entries"/>). In the case of an <spanx style="verb">x509_entry_v2</spanx> entry, this is the whole <spanx style="verb">X509ChainEntry</spanx>; and in the case of a <spanx style="verb">precert_entry_v2</spanx>, this is the whole <spanx style="verb">PrecertChainEntryV2</spanx>.
                     </t>
                     <t hangText="sct:">
-                      A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">x509_sct_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> corresponding to this log entry. Note that more than one SCT may have been returned for the same entry - only one of those is returned in this field. It may not be possible to retrieve others.
+                      A base64 encoded <spanx style="verb">TransItem</spanx> of type <spanx style="verb">x509_sct_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> corresponding to this log entry.
```
benlaurie: s/A/The/

PR #160, opened 2016-05-22
Clarity - "consists of" -> "contains"

wkumari:
"Each log consists of certificate chains" -> "Each log contains certificate chains..." 

"consists of" makes it sound like certificate chains are the only thing in the log.
I'm suggesting "contains", but I'm not sure it is much better. 
Feel free to reject / ignore.

Apologies,  diff not highlighting changes again.


wkumari:
Submitted with incorrect "Primary email" selected.

PR #167, opened 2016-05-27
Removing duplicate text.

eranmes:
That section exists twice, remove it from the non-optional part.


eranmes:
@benlaurie PTAL because I suspect your merge shenanigans...

benlaurie:
Highly likely. Sorry about that.

PR #168, opened 2016-05-27
Section 3.1: Clarify the "may" is not permissive.

wkumari:
"may" sounds permissive, "might" sounds better; otherwise this sounds like a permission and not a prediction.
Alternatively, the proposed text...

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -307,7 +307,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       </t>
       <section title="Certificates">
         <t>
-          Any entity can <xref target="add-chain">submit a certificate</xref> to a log. Since certificates may not be accepted by TLS clients unless logged, it is expected that certificate owners or their CAs will usually submit them.
+          Any entity can <xref target="add-chain">submit a certificate</xref> to a log. It is anticipated that TLS clients will not accept certificates which are not logged, and so it is expected that certificate owners (or their CAs) will submit them.
         </t>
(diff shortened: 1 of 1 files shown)


eranmes:
@benlaurie PTAL.

rfc6962-bis.xml
```diff
@@ -307,7 +307,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       </t>
       <section title="Certificates">
         <t>
-          Any entity can <xref target="add-chain">submit a certificate</xref> to a log. Since certificates may not be accepted by TLS clients unless logged, it is expected that certificate owners or their CAs will usually submit them.
+          Any entity can <xref target="add-chain">submit a certificate</xref> to a log. It is anticipated that TLS clients will not accept certificates which are not logged, and so it is expected that certificate owners (or their CAs) will submit them.
```
robstradling: s/not accept certificates which/reject certificates that/

Why add brackets around "or their CAs"?  Today it's far more common for CAs to submit certs to logs than it is for certificate owners to submit them.  "certificate owners or their CAs" does sound a bit clumsy though.

How about this?
"Since it is anticipated that TLS clients will reject certificates that are not logged, it is expected that certificate issuers and subjects will be strongly motivated to submit them."

benlaurie: Agree with Rob's comments. OTOH, don't see any particular need to change this, not sure what the proposal was supposed to fix.

eranmes:
LGTM, will merge now.

PR #173, opened 2016-06-15
Clarify the role of the TLS feature extension

AGWA:
Addresses my comments on Sections 10.2.5 - 10.2.7: https://mailarchive.ietf.org/arch/msg/trans/mhQg2dYwygll-yStm9MbTCFxxt8

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1460,9 +1460,6 @@ but it is expected there will be a variety.
           <t>
             If any certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension, then CT compliance (using any of the mechanisms from <xref target="tls_servers"/>) is required.
           </t>
-          <t>
-            TLS clients MUST treat certificates which fail this requirement as non-compliant.
-          </t>
         </section>
         <section title="Handling of Non-compliance">
           <t>


rfc6962-bis.xml
```diff
@@ -1468,10 +1465,10 @@ but it is expected there will be a variety.
           <t>
             <list style="symbols">
               <t>
-                In the case that use of TLS with a valid certificate is mandated by explicit security policy, application protocol specification, or other means, the TLS client MUST refuse the connection.
+                In the case that use of TLS with a valid certificate is mandated by explicit security policy, application protocol specification, or other means, the TLS client SHOULD refuse the connection.
```
eranmes: Why do you suggest changing the MUST to SHOULD here and below?

rfc6962-bis.xml
```diff
@@ -1455,10 +1455,7 @@ but it is expected there will be a variety.
         </section>
         <section title="TLS Feature Extension">
           <t>
-            If any certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension, then CT compliance (using any of the mechanisms from <xref target="tls_servers"/>) is required.
-          </t>
-          <t>
-            TLS clients MUST treat certificates which fail this requirement as non-compliant.
+            If any certificate in a chain includes the <xref target="tls_transinfo_extension">transparency_info</xref> TLS extension identifier in the <xref target="RFC7633">TLS Feature</xref> certificate extension, then CT compliance (using any of the mechanisms from <xref target="tls_servers"/>) is required, and TLS clients MUST refuse the connection if the chain is not compliant.
```
eranmes: FYI that change looks good.

robstradling: I agree that the "TLS clients MUST treat certificates which fail this requirement as non-compliant" sentence is redundant, because we already say "...then CT compliance...is required".

However, the "TLS clients MUST refuse the connection" part belongs in the "Handling of Non-compliance" section, not here.

rfc6962-bis.xml
robstradling: We liked https://tools.ietf.org/html/rfc7633#section-4.3.3 so we copied'n'pasted it.  :-)

It's a conditional MUST, not an absolute MUST.  When the "In the case that...mandated by..." condition is met, MUST is appropriate.  (IIRC we wanted to avoid absolute MUSTs/SHOULDs, because specifying client behaviour is out of scope for 6962-bis).

How about we change the 2 occurrences of "(the) use of TLS with a valid certificate" to "compliance" ?

AGWA: > How about we change the 2 occurrences of "(the) use of TLS with a valid certificate" to "compliance" ?

I don't think that makes sense, because then it's saying that if compliance is optional (e.g. there's no TLS feature extension), clients MUST NOT treat the certificate as valid if it's non-compliant.

How about making the two bullet points conditional on "If CT compliance is required"?  I just pushed a new commit with proposed wording.

eranmes:
LGTM, @robstradling what do you think?

@AGWA Can you please rebase, there seem to be some merge conflicts?

AGWA:
Rebased.  It turns out the changes for Stephen Kent's feedback (7ffc0ba87c33ad511c8583f6df673815f605d5cc) already addressed most of my concerns, so this diff is now very short.

eranmes:
Thanks, could you please make sure you sign the CLA?

robstradling:
LGTM too.

PR #174, opened 2016-06-21
Addressing feedback from Steve Kent.

eranmes:
See http://www.ietf.org/mail-archive/web/trans/current/msg02308.html

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -338,7 +338,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
     </section>
     <section title="Private Domain Name Labels">
       <t>
-        Some regard some DNS domain name labels within their registered domain space as private and security sensitive. Even though these domains are often only accessible within the domain owner's private network, it's common for them to be secured using publicly trusted TLS server certificates. We define a mechanism to allow these private labels to not appear in public logs.
+        Some regard some DNS domain name labels within their registered domain space as private and security sensitive. Even though these domains are often only accessible within the domain owner's private network, it's common for them to be secured using publicly trusted TLS server certificates. We define a mechanism to allow these private labels to not appear in public logs, while still retaining most of the security benefits that accrue from using Certificate Transparency mechanisms.
       </t>
       <section title="Wildcard Certificates">
         <t>
@@ -368,7 +368,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       </section>
       <section title="Using a Name-Constrained Intermediate CA" anchor="name_constrained">
         <t>
-          An intermediate CA certificate or intermediate CA precertificate that contains the critical or non-critical <xref target="RFC5280">Name Constraints</xref> extension MAY be logged in place of end-entity certificates issued by that intermediate CA, as long as all of the following conditions are met:
+          An intermediate CA certificate or intermediate CA precertificate that contains the <xref target="RFC5280">Name Constraints</xref> extension MAY be logged in place of end-entity certificates issued by that intermediate CA, as long as all of the following conditions are met:
           <list style="symbols">
             <t>
               there MUST be a non-critical extension (OID 1.3.101.76, whose extnValue OCTET STRING contains ASN.1 NULL data (0x05 0x00)). This extension is an explicit indication that it is acceptable to not log certificates issued by this intermediate CA.
@@ -474,7 +474,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Log ID" anchor="log_id">
         <figure>
           <preamble>
-            Each log is uniquely identified by an OID (specified in the log's metadata). A log's operator MUST either allocate the OID themselves or request an OID from one of the two Log ID Registries (see <xref target="log_id_registry1"/> and <xref target="log_id_registry2"/>). Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
+            Each log is identified by an OID, which is specified in the log's metadata and which MUST NOT be used to identify any other log. A log's operator MUST either allocate the OID themselves or request an OID from one of the two Log ID Registries (see <xref target="log_id_registry1"/> and <xref target="log_id_registry2"/>). Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
           </preamble>
           <artwork>
     opaque LogID&lt;2..127&gt;;</artwork>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -300,7 +300,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
     </section>
     <section title="Submitters">
       <t>
-        Submitters submit certificates or preannouncements of certificates prior to issuance (precertificates) to logs for public auditing, as described below. In order to enable attribution of each logged certificate or precertificate to its issuer, each submission MUST be accompanied by all additional certificates required to verify the chain up to an accepted trust anchor. The trust anchor (a root or intermediate CA certificate) MAY be omitted from the submission.
+        Submitters submit certificates or pre-registrations of certificates prior to issuance (precertificates) to logs for public auditing, as described below. In order to enable attribution of each logged certificate or precertificate to its issuer, each submission MUST be accompanied by all additional certificates required to verify the chain up to an accepted trust anchor. The trust anchor (a root or intermediate CA certificate) MAY be omitted from the submission.
```
robstradling: I prefer the term "preannouncements".  We don't talk about submission equating to any sort of "registration" anywhere else in the document.

http://www.thefreedictionary.com/preannounce "to announce in advance"
http://www.thefreedictionary.com/announce "To make known publicly"

So a preannouncement makes a thing known publicly in advance.  That's exactly what a precertificate does, isn't it?

rfc6962-bis.xml
```diff
@@ -312,7 +312,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          CAs may pre-register a certificate prior to issuance by <xref target="add-pre-chain">submitting a precertificate</xref> that the log can use to create an entry that will be valid against the issued certificate. The CA MAY incorporate the returned SCT in the issued certificate. Examples of situations where the returned SCT is not incorporated into the issued certificate would be when a CA sends the precertificate to multiple logs, but only incorporates the SCTs that are returned first, or the CA is using <xref target="redacting_labels">domain name redaction</xref> and intends to use another mechanism to publish SCTs (such as an <xref target="ocsp_transinfo_extension">OCSP response</xref> or the <xref target="tls_transinfo_extension">TLS extension</xref>).
```
robstradling: Again, I prefer "preannounce".

rfc6962-bis.xml
```diff
@@ -338,7 +338,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
     </section>
     <section title="Private Domain Name Labels">
       <t>
-        Some regard some DNS domain name labels within their registered domain space as private and security sensitive. Even though these domains are often only accessible within the domain owner's private network, it's common for them to be secured using publicly trusted TLS server certificates. We define a mechanism to allow these private labels to not appear in public logs.
+        Some regard some DNS domain name labels within their registered domain space as private and security sensitive. Even though these domains are often only accessible within the domain owner's private network, it's common for them to be secured using publicly trusted TLS server certificates. We define a mechanism to allow these private labels to not appear in public logs, while still retaining some of the security benefits that accure from using Certificate Transparency mechanisms.
```
robstradling: s/accure/accrue/

Would it be fair to change "some" to "most" or "many" ?

rfc6962-bis.xml
```diff
@@ -474,7 +474,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Log ID" anchor="log_id">
         <figure>
           <preamble>
-            Each log is uniquely identified by an OID (specified in the log's metadata). A log's operator MUST either allocate the OID themselves or request an OID from one of the two Log ID Registries (see <xref target="log_id_registry1"/> and <xref target="log_id_registry2"/>). Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
+            Each log is identified by an OID (specified in the log's metadata) which MUST be unique. A log's operator MUST either allocate the OID themselves or request an OID from one of the two Log ID Registries (see <xref target="log_id_registry1"/> and <xref target="log_id_registry2"/>). Various data structures include the DER encoding of this OID, excluding the ASN.1 tag and length bytes, in an opaque vector:
```
robstradling: IIUC, Steve's comment was that he wants the "allocate the OID themselves" option to disappear, because he foresees Log ID collisions if this option is permitted.  I don't share his paranoia.  An organization can only allocate an OID under an OID arc that they own.  If a log operator can't manage their own OID arc properly, then I think they'd probably be just as likely to make the mistake of requesting ONE OID from one of the Log ID Registries and then using that ONE OID for TWO logs.

robstradling: I prefer the original "uniquely identified by an OID" text.

"an OID...which MUST be unique" doesn't really say anything.  Each OID is unique by definition!  e.g. there is only 1 OID that has the value 1.2.3.4.

The important point we need to make is that each OID MUST only be used as the Log ID for 1 log.  I think "is uniquely identified" expresses this.  That said, if you feel that it needs to be expressed more verbosely, how about this...

"Each log is identified by an OID, which is specified in the log's metadata and which MUST NOT be used to identify any other log."

rfc6962-bis.xml
```diff
@@ -1319,7 +1319,7 @@ entry specified by <spanx style="verb">start</spanx>.
...
+            The Transparency Information X.509v3 extension, which has OID 1.3.101.75 and MUST be non-critical, contains one or more <spanx style="verb">TransItem</spanx> structures in a <spanx style="verb">TransItemList</spanx>. This extension MAY be included in OCSP responses (see <xref target="ocsp_transinfo_extension"/>) and certificates (see <xref target="cert_transinfo_extension"/>). Since RFC5280 requires the <spanx style="verb">extnValue</spanx> field (an OCTET STRING) of each X.509v3 extension to include the DER encoding of an ASN.1 value, a <spanx style="verb">TransItemList</spanx> MUST NOT be included directly. Instead, it MUST be wrapped inside an additional OCTET STRING, which is then put into the <spanx style="verb">extnValue</spanx> field:
```
robstradling: Is it conceivable that a CA might sometimes want to issue certs that aren't backwards-compatible with clients that don't support CT?  If so, I prefer "SHOULD".

rfc6962-bis.xml
eranmes: Reverted.

rfc6962-bis.xml
eranmes: Reverted.

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: Switched to your suggested text.
Don't share the concern about operators allocating OIDs themselves.

rfc6962-bis.xml
eranmes: Which of the MUSTs are you referring to?

eranmes:
@robstradling PTAL

rfc6962-bis.xml
robstradling: I'm referring to the one-word change that this diff makes to this line, and suggesting that there might be a good reason to reject this one-word change.

If a CA might ever want to issue a cert that is deliberately _not_ backwards-compatible with clients that don't support CT, then I think we should permit the Transparency Information extension to be critical.

eranmes: Understood - I agree we should leave this choice to CAs.

eranmes:
Any other comments @robstradling ?

robstradling:
LGTM

PR #175, opened 2016-06-21
First stab at simplifed redaction.

eranmes:
I'm sure it can be articulated in a better way.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Redaction of Domain Name Labels" anchor="redacting_labels">
         <section title="Redacting Labels in Precertificates">
           <t>
-            When creating a precertificate, the CA MAY substitute one or more labels in each DNS-ID and <xref target="RFC6125">CN-ID</xref> with a corresponding number of <spanx style="verb">?</spanx> labels. Every label to the left of a <spanx style="verb">?</spanx> label MUST also be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">?.?.example.com</spanx> instead, but not <spanx style="verb">top.?.example.com</spanx> instead.
+            When creating a precertificate, the CA MAY omit the <xref target="RFC5280">subjectAltName</xref> extension. Instead, the CA MUST include a <xref target="redacted_san_extension">Redacted SAN</xref> extension, containing the entries that will be included in the certificate's subjectAltName extension, in a redacted form.
           </t>
           <t>
-            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.?.?.example.com</spanx> instead, but not <spanx style="verb">?.?.?.example.com</spanx> instead.
+            One or more labels in each <xref target="RFC6125">DNS-ID</xref> can be replaced with <spanx style="verb">REDACTED(label)</spanx> labels.
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || BASE32(HASH_INDEX || HASH(subject_key_identifier || label))
+            Where || denotes concatenation, HASH_INDEX is the index of a hash function in <xref target="hash_algorithms"/>, HASH is the hash function identified by HASH_INDEX, subject_key_identifier is the content bytes of the Subject Key Identifier extension (section 4.2.1.2. of <xref target="RFC5280"/>) and BASE32 is base32 encoding of the value according to <xref target="RFC4648"/>, section 6, excluding the padding.
+          </t>
+          <t>
+            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted.
           </t>
         </section>
-        <section title="Redacted Labels Certificate Extension" anchor="redaction_extension">
+        <section title="Redacted SAN Certificate Extension" anchor="redacted_san_extension">
           <t>
-            When a precertificate contains one or more <spanx style="verb">?</spanx> labels, a non-critical extension (OID 1.3.101.77, whose extnValue OCTET STRING contains an ASN.1 SEQUENCE OF INTEGERs) MUST be added to the corresponding certificate. The purpose of this extension is to enable TLS clients to reconstruct the TBSCertificate component of the precertificate from the certificate, as described in <xref target="reconstructing_tbscertificate"/>.
+            The Redacted SAN extension is a non-critical extension (OID 1.3.101.77) identical in structure to the subjectAltName extension, the difference being that dNSName entries MAY contain redacted labels. The purpose of this extension is to inform TLS clients of the dNSNames that were redacted and the degree of redaction, while minimizing the complexity of TBSCertificate reconstruction, as described in <xref target="reconstructing_tbscertificate"/>.
           </t>
           <t>
-            Each INTEGER MUST have a value of zero or more. The first INTEGER indicates the total number of <spanx style="verb">?</spanx> labels in the precertificate's first DNS-ID; the second INTEGER does the same for the precertificate's second DNS-ID; etc. The last INTEGER does the same for the precertificate's zero or more CN-IDs. There MUST NOT be more INTEGERs than there are DNS-IDs (plus one, if any CN-IDs are present); if there are fewer INTEGERs than this, the shortfall is made up by implicitly repeating the last INTEGER.
+            When used, the Redacted SAN extension MUST be present in both the precertificate and the corresponding certificate.
+          </t>
+          <t>
+            Hashing the redacted labels allows the legitimate domain owner to identify if the redacted label correlates to one of the labels they're aware of or a new, potentially unauthorized, subdomain.
           </t>
         </section>
       </section>
@@ -1412,21 +1424,39 @@ but it is expected there will be a variety.
             To reconstruct the TBSCertificate component of a precertificate from a certificate, TLS clients should:
             <list style="symbols">
               <t>
-                Remove the Transparency Information extension described in <xref target="x509v3_transinfo_extension"/>
+                Remove the Transparency Information extension described in <xref target="x509v3_transinfo_extension"/>.
               </t>
               <t>
-                Remove the Redacted Labels extension described in <xref target="redaction_extension"/>
-              </t>
-              <t>
-                Replace leftmost labels of each DNS-ID with <spanx style="verb">?</spanx>, based on the INTEGER value corresponding to that DNS-ID in the extension.
+                If the Redacted SAN extension (<xref target="redacted_san_extension"/>) is present:
+                <list style="symbols">
+                  <t>TLS clients MUST verify it against the subjectAltName extension according to <xref target="verifying_redacted_san"/>.</t>
+                  <t>Once verified, remove the subjectAltName extension from the TBSCertificate.</t>
+                </list>
               </t>
             </list>
           </t>
           <t>
-            A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
+            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_sct_v2</spanx>), then the client SHOULD also remove embedded v1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded v1 and v2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
+          </t>
+        </section>
+        <section title="Verifying the Redacted SAN extension" anchor="verifying_redacted_san">
+          <t>
+            If the Redacted SAN extension is present, TLS clients MUST check that the SAN extension is present, that the SAN extension contains the same number of identities as the Redacted SAN extension, and that each identity in the SAN extension has a matching identity at the same position in the Redacted SAN extension. Two identities are matching if either:
+            <list style="symbols">
+              <t>The two identities are identical; or,</t>
+              <t>Both identities are DNS names, have the same number of labels, and each label in the SAN identity has a matching label at the same position in the Redacted SAN identity. Two DNS labels are matching if either:
+                <list style="symbols">
+                  <t>The two labels are identical; or,</t>
+                  <t>The label from the the Redacted SAN identity is equal to REDACTED(label from SAN identity) (<xref target="redacting_labels"/>)</t>
+                </list>
+              </t>
+            </list>
           </t>
           <t>
-            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_sct_v2</spanx>), then the client SHOULD also remove embedded v1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded v1 and v2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
+            If any of these checks fail, the certificate MUST NOT be considered compliant.
+          </t>
+          <t>
+            A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
           </t>
         </section>
         <section title="Validating SCTs">
@@ -1714,7 +1744,7 @@ but it is expected there will be a variety.
       </section>
       <section title="Object Identifiers">
         <t>
-          This document uses object identifiers (OIDs) to identify Log IDs (see <xref target="log_id"/>), the precertificate CMS <spanx style="verb">eContentType</spanx> (see <xref target="Precertificates"/>), and X.509v3 extensions in certificates (see <xref target="redaction_extension"/>, <xref target="name_constrained"/> and <xref target="cert_transinfo_extension"/>) and OCSP responses (see <xref target="ocsp_transinfo_extension"/>). The OIDs are defined in an arc that was selected due to its short encoding.
+          This document uses object identifiers (OIDs) to identify Log IDs (see <xref target="log_id"/>), the precertificate CMS <spanx style="verb">eContentType</spanx> (see <xref target="Precertificates"/>), and X.509v3 extensions in certificates (see <xref target="redacted_san_extension"/>, <xref target="name_constrained"/> and <xref target="cert_transinfo_extension"/>) and OCSP responses (see <xref target="ocsp_transinfo_extension"/>). The OIDs are defined in an arc that was selected due to its short encoding.
         </t>
         <section title="Log ID Registry 1" anchor="log_id_registry1">
           <t>
@@ -1764,7 +1794,7 @@ but it is expected there will be a variety.
       </section>
       <section title="Avoiding Overly Redacting Domain Name Labels">
         <t>
-          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. <spanx style="verb">?.com</spanx>, <spanx style="verb">?.co.uk</spanx> and other <xref target="Public.Suffix.List">public suffixes</xref>), then the domain name may be considered by clients to be overly redacted.
+          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. <spanx style="verb">REDACTED(label).com</spanx>, <spanx style="verb">REDACTED(label).co.uk</spanx> and other <xref target="Public.Suffix.List">public suffixes</xref>), then the domain name may be considered by clients to be overly redacted.
         </t>
         <t>
           CAs should take care to avoid overly redacting domain names in precertificates. It is expected that monitors will treat precertificates that contain overly redacted domain names as potentially misissued. TLS clients MAY consider a certificate to be non-compliant if the <xref target="reconstructing_tbscertificate">reconstructed TBSCertificate</xref> contains any overly redacted domain names.


rfc6962-bis.xml
```diff
@@ -351,18 +351,33 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
           </t>
           <t>
-            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.?.?.example.com</spanx> instead, but not <spanx style="verb">?.?.?.example.com</spanx> instead.
+            One or more labels in each <xref target="RFC6125">DNS-ID</xref> can be replaced with <spanx style="verb">REDACTED(label)</spanx> labels.
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || HEX(HASH(serial_number || label))
+            Where || denotes concatenation, HASH is the hash function indexed 0 in <xref target="hash_algorithms"/>, serial_number is the precertificate's serial number, consisting of the ASN.1 INTEGER tag, length and content bytes.
```
robstradling: "HASH is the hash function indexed 0" lacks algorithm agility.

The CA already has to use a hash function when signing the precertificate and corresponding certificate.  Why don't we say that the same hash function MUST be used here?

The client would need a Signature Algorithm -> Hash Algorithm lookup table, but I think it'd be worth it for the agility.

eranmes: I agree the current text lacks algorithm agility.
I'm not convinced the complexity is worth the benefit of being able to use different hash algorithms, compared to the interoperability problems that may be caused by allowing different hash algorithms.

eranmes: Per our offline discussion, the drawback of specifying a single hash is that it's always required to be present in the code for this purpose even if it's obsolete for any other purpose. So using the hash algorithm used by the certificate sounds reasonable.

eranmes:
@robstradling PTAL.

rfc6962-bis.xml
```diff
@@ -351,18 +351,33 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
-            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.?.?.example.com</spanx> instead, but not <spanx style="verb">?.?.?.example.com</spanx> instead.
+            One or more labels in each <xref target="RFC6125">DNS-ID</xref> can be replaced with <spanx style="verb">REDACTED(label)</spanx> labels.
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || HEX(HASH(serial_number || label))
+            Where || denotes concatenation, HASH is the hash function from the TBSCertificate's signature field (<xref target="RFC5280">RFC5280, Section 4.1.2.3.</xref>), serial_number is the precertificate's serial number, consisting of the ASN.1 INTEGER tag, length and content bytes.
```
benlaurie: a) I think you should refer to the signaureAlgorithm field.

b) more seriously, there's no requirement that a signature uses a hash at all.

c) HASH is already defined by the log - use that?

rfc6962-bis.xml
```diff
@@ -351,18 +351,33 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          </t>
+          <t>
+            Every label to the left of a redacted label MUST also be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">REDACTED(top).REDACTED(secret).example.com</spanx> instead, but not <spanx style="verb">top.REDACTED(secret).example.com</spanx> instead.
```
benlaurie: Not clear why this is required? Doesn't seem to add any security, nor any noticeable simplicity.

rfc6962-bis.xml
```diff
@@ -351,18 +351,33 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          </t>
+          <t>
+            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.REDACTED(top).REDACTED(secret).example.com</spanx> instead, but not <spanx style="verb">REDACTED(*).REDACTED(top).REDACTED(secret).example.com</spanx> instead.
```
benlaurie: Again, why not redact wildcards?

rfc6962-bis.xml
```diff
@@ -1415,10 +1430,7 @@ but it is expected there will be a variety.
                 Remove the Transparency Information extension described in <xref target="x509v3_transinfo_extension"/>
               </t>
               <t>
-                Remove the Redacted Labels extension described in <xref target="redaction_extension"/>
-              </t>
-              <t>
-                Replace leftmost labels of each DNS-ID with <spanx style="verb">?</spanx>, based on the INTEGER value corresponding to that DNS-ID in the extension.
+                If the <xref target="redacted_san_extension">Redacted SAN</xref> extension is present, remove the subjectAltName extension.
```
benlaurie: Surely needs to be a description of verifying the redacted SANs?

rfc6962-bis.xml
AGWA: @benlaurie Using the log's HASH doesn't work, since a single pre-certificate might be submitted to several different logs, each of which may use a different hash function.

Perhaps the hash algorithm index could be specified in the redacted label, like this:

```
REDACTED(label) = ? || HASH_INDEX || ? || HEX(HASH(serial_number || label))
```

e.g.:

```
?0?e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

AGWA: The case of HEX needs to be defined.

rfc6962-bis.xml
AGWA: I agree there should be a description.  I also think that at least some amount of verification should be required of clients.  We don't want to allow a pre-certificate for `?.example.com` to be used with `www.example.org` for instance.

Here's some suggested text:

> If the Redacted SAN extension is present, TLS clients MUST check that the SAN extension is present, that the SAN extension contains the same number of identifiers as the Redacted SAN extension, and that each identifier in the SAN extension has a matching identifier at the same position in the Redacted SAN extension.  Two identifiers are matching if either:
>
> * The two identifiers are identical; or,
>
> * Both identifiers are DNS names, have the same number of labels, and each label in the SAN identifier has a matching label at the same position in the Redacted SAN identifier.  Two DNS labels are matching if either:
>
>   * The two labels are identical; or,
>
>   * The label from the the Redacted SAN identifier is equal to REDACT(label from SAN identifier) [cross-reference to redacting_labels]
>
> If any of these checks fail, the certificate MUST NOT be considered compliant.

rfc6962-bis.xml
eranmes: Not necessary, removed.

rfc6962-bis.xml
eranmes: Even though it's technically possible, it's already a form of redaction. 

Under the current proposal, whoever has a list of all the subdomains under the domain where redaction takes place (i.e. everything under example.com), can verify if it is one of the subdomains it knows about that was redacted or not.

By allowing redaction of a wildcard, we effectively allow hiding the name that was redacted without the option to 'un'-redact it. It is (mostly) equivalent to the current text that allows redacting using question marks only.

rfc6962-bis.xml
eranmes: @benlaurie good point about the signatureAlgorithm not requiring to use a hash at all. @robstradling , as you originally suggested using that, any thoughts?

@AGWA - why do you think the case of the hex needs to be defined?

eranmes:
@benlaurie and @AGWA PTAL.

rfc6962-bis.xml
eranmes: Adopted.

rfc6962-bis.xml
AGWA: @eranmes - It would be helpful for clients to know whether to expect lower or uppercase, and for CAs to know what case to use.  If you want to leave the case unspecified, I think it would be valuable to explicitly state this so clients know they need to be prepared for both, in case some assume that it's always a particular case because they've only seen certs using that case so far.

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || HEX(HASH(serial_number || label))
+            Where || denotes concatenation, HASH is the hash function from the TBSCertificate's signature field (<xref target="RFC5280">RFC5280, Section 4.1.2.3.</xref>), serial_number is the precertificate's serial number, consisting of the ASN.1 INTEGER tag, length and content bytes.
+          </t>
+          <t>
+            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.REDACTED(top).REDACTED(secret).example.com</spanx> instead, but not <spanx style="verb">REDACTED(*).REDACTED(top).REDACTED(secret).example.com</spanx> instead.
```
AGWA: This paragraph can be reduced to just the first sentence, since there is no longer a requirement that all left-most labels be redacted.

rfc6962-bis.xml
robstradling: @eranm @benlaurie Yeah, that's a fair point about sigalgs not being required to use hashes.  I like @AGWA's "? || HASH_INDEX || ?" proposal.

@AGWA I'd say it's **vital** for clients that the case of HEX is defined.  The Redacted SAN extension is present in the TBSCertificate that the client needs to reconstruct, so there is no room for ambiguity.  (If left undefined, what's to stop a CA from including several kilobytes of MiXEd-cASe HEX redacted labels?!)

robstradling: "consisting of the ASN.1 INTEGER..." should probably be "consisting of the DER encoding of the ASN.1 INTEGER..."

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
           </t>
         </section>
-        <section title="Redacted Labels Certificate Extension" anchor="redaction_extension">
+        <section title="Redacted SAN Certificate Extension" anchor="redacted_san_extension">
+          <t>
+            The Redacted SAN extension is a non-critical extension (OID 1.3.101.77) identical in structure to the subjectAltName extension, the difference being that dNSName entries MAY contain redacted labels. The purpose of this extension is to inform TLS clients of the dNSNames that were redacted and the degree of redaction, while minimizing the complexity of TBSCertificate reconstruction, as described in <xref target="reconstructing_tbscertificate"/>.
```
robstradling: @eranmes Given that Symantec have implemented the redaction mechanism in -16, perhaps we shouldn't "reuse" the 1.3.101.77 OID for the new Redacted SAN extension.  (It would be a shame to lose that particular OID though: it's your easter egg ;-) ).

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+          </t>
           <t>
-            When a precertificate contains one or more <spanx style="verb">?</spanx> labels, a non-critical extension (OID 1.3.101.77, whose extnValue OCTET STRING contains an ASN.1 SEQUENCE OF INTEGERs) MUST be added to the corresponding certificate. The purpose of this extension is to enable TLS clients to reconstruct the TBSCertificate component of the precertificate from the certificate, as described in <xref target="reconstructing_tbscertificate"/>.
+            The Redacted SAN extension MUST be present in the certificate corresponding to the precertificate that was logged with it.
```
robstradling: Can I suggest rephrasing this sentence as follows:
"When used, the Redacted SAN extension MUST be present in both the precertificate and the corresponding certificate."

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
           </t>
           <t>
-            Each INTEGER MUST have a value of zero or more. The first INTEGER indicates the total number of <spanx style="verb">?</spanx> labels in the precertificate's first DNS-ID; the second INTEGER does the same for the precertificate's second DNS-ID; etc. The last INTEGER does the same for the precertificate's zero or more CN-IDs. There MUST NOT be more INTEGERs than there are DNS-IDs (plus one, if any CN-IDs are present); if there are fewer INTEGERs than this, the shortfall is made up by implicitly repeating the last INTEGER.
+            The purpose behind hashing the redacted labels is to allow the legitimate domain owner to identify if the redacted precertificate correlates to one of the subdomains it's aware of or a new, potentially unautorized, subdomain.
```
robstradling: s/it's/they're/

s/unautorized/unauthorized/

robstradling: "if the redacted precertificate correlates to one of the subdomains" seems imprecise.
How about "if the redacted label correlates to one of the labels" ?

Also, how about explaining **how** the legitimate domain owner is supposed to perform that identification process.  i.e. they need to calculate REDACTED(label) for each label they know about, then check if that found a match.

rfc6962-bis.xml
```diff
@@ -1412,21 +1424,39 @@ but it is expected there will be a variety.
...
+            </list>
+          </t>
+          <t>
+            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_sct_v2</spanx>), then the client SHOULD also remove embedded v1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded v1 and v2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
+          </t>
+        </section>
+        <section title="Verifying the Redacted SAN extension" anchor="verifying_redacted_san">
+          <t>
+            If the Redacted SAN extension is present, TLS clients MUST check that the SAN extension is present, that the SAN extension contains the same number of identifiers as the Redacted SAN extension, and that each identifier in the SAN extension has a matching identifier at the same position in the Redacted SAN extension. Two identifiers are matching if either:
```
robstradling: "same number of identifiers...matching identifier at the same position".  Shouldn't this say "same number of dNSNames...matching label at the same position"?

AGWA: >  "same number of identifiers...matching identifier at the same position". Shouldn't this say "same number of dNSNames...matching label at the same position"?

No, that's the right wording.  dNSName matching is described a few lines down, and it's vital that non-dNSName SANs be checked so the CA can't omit them from the pre-certificate to evade monitoring.

robstradling: Ah, I see.  (I really should stop trying to work so late in the evening ;-) ).

Could we sync the terminology to RFC5280?  5280 Section 4.2.1.6 talks about "identities" rather than "identifiers".  Also, I wonder if "same position in the Redacted SAN extension" could be more precisely expressed in terms of "GeneralName(s)".

rfc6962-bis.xml
robstradling: Ignore my "it's vital for clients that the case of HEX is defined".  I forgot that the Redacted SAN is present in the corresponding certificate, and so the client doesn't need to reconstruct it.

rfc6962-bis.xml
eranmes: As discussed offline, there's no chance for confusion because Symantec's (hasty!) use of the 1.3.101.77 OID is only found in 6962 precertificates, not in 6962-bis CMS precertificates, so the context indicates how to interpret it.

rfc6962-bis.xml
eranmes: (1) changed precertificate -> label.
(2) I think the *how* is straightforward (if the reader was able to follow generation of redacted labels, they could figure out how to find out a match between a redacted label and a potential source label)

rfc6962-bis.xml

rfc6962-bis.xml
eranmes: (1) terminology synced.
(2) Can you suggest a concrete change? I don't see a problem with the current wording, where each identity has to be either identical or a dNSName with labels that are identical or redacted.

eranmes:
@AGWA , @robstradling  and @benlaurie PTAL.

rfc6962-bis.xml
```diff
@@ -1412,21 +1424,39 @@ but it is expected there will be a variety.
...
+        </section>
+        <section title="Verifying the Redacted SAN extension" anchor="verifying_redacted_san">
+          <t>
+            If the Redacted SAN extension is present, TLS clients MUST check that the SAN extension is present, that the SAN extension contains the same number of identities as the Redacted SAN extension, and that each identity in the SAN extension has a matching identity at the same position in the Redacted SAN extension. Two identities are matching if either:
+            <list style="symbols">
+              <t>The two identities are identical; or,</t>
+              <t>Both identities are DNS names, have the same number of labels, and each label in the SAN identity has a matching label at the same position in the Redacted SAN identity. Two DNS labels are matching if either:
+                <list style="symbols">
+                  <t>The two labels are identical; or,</t>
+                  <t>The label from the the Redacted SAN identifier is equal to REDACTED(label from SAN identifier) (<xref target="redacting_labels"/>)</t>
```
AGWA: s/identifier/identity/g

AGWA:
@eranmes Based on your experience with TLS clients, do you think using the "DER encoding of the ASN.1 INTEGER tag, length and content bytes" as the salt will pose any challenges to clients?  Either their parsers will need to provide access to the original DER encoding of the serial number, or they'll need to re-encode it (which would produce different results if the serial number was not minimally encoded).

robstradling:
@AGWA @eranmes Two ideas for avoiding the INTEGER encoding issues:

(1) Use "HASH(SubjectPublicKeyInfo)" instead of "serial number".  SPKI is less likely to be unique (because multiple certs can share the same public key), but I think it's probably "salty" enough.  However, we might run into problems with incorrectly encoded signature parameters though.

(2) Use the contents bytes of the Subject Key Identifier extension instead of "serial number".  I'm guessing this might be easier (than "HASH(SubjectPublicKeyInfo)") for CAs and clients to implement.  RFC5280 only says that the SKI extension SHOULD be included, but I suppose we could say it MUST be included if you're doing redaction.

Of these two ideas, I prefer (2).

What do you think?

rfc6962-bis.xml

eranmes:
@AGWA Exposing the original DER encoding of the serial number may require some changes in libraries clients use for parsing the certificate, but that is much less of a challenge than requiring clients to re-encode the DER, as certificate parsers, in their minimal form, don't necessarily possess the ability to encode DER (as I've mentioned on the list, I believe).

@robstradling I don't have a strong opinion on using the SKI, it may be slightly simpler to implement as the X.509 parsers I know have an API for extracting extensions' contents relatively easily.

eranmes:
Follow up on simplicity of implementation:
Clients using BoringSSL/OpenSSL would have to re-encode the serial number as it's available in the form of an ASN.1 integer:
https://github.com/openssl/openssl/blob/master/include/openssl/x509.h#L620
(or in boringssl: https://github.com/google/boringssl/blob/master/include/openssl/x509.h#L214)

Whereas the SKI is available as an octet string easily:
https://github.com/openssl/openssl/blob/master/include/openssl/x509v3.h#L661
(or in boringssl: https://github.com/google/boringssl/blob/master/include/openssl/x509.h#L257)

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
           </t>
           <t>
-            Wildcard <spanx style="verb">*</spanx> labels MUST NOT be redacted. However, if the complete leftmost label of a DNS-ID or CN-ID is <spanx style="verb">*</spanx>, it is considered redacted for the purposes of determining if the label to the right may be redacted. For example, if a certificate contains a DNS-ID of <spanx style="verb">*.top.secret.example.com</spanx>, then the corresponding DNS-ID in the precertificate could contain <spanx style="verb">*.?.?.example.com</spanx> instead, but not <spanx style="verb">?.?.?.example.com</spanx> instead.
+            One or more labels in each <xref target="RFC6125">DNS-ID</xref> can be replaced with <spanx style="verb">REDACTED(label)</spanx> labels.
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || HASH_INDEX || ? || HEX(HASH(serial_number || label))
```
benlaurie: Hmm, not sure I like the ? bit - why not define the HASH_INDEX as a single byte (0xff reserved for expansion)?

robstradling: @benlaurie Which ? don't you like?  The first?  The second?  Both?

HASH_INDEX is the string representation of the hash index.  You can't put a literal 0xff in an ASN.1 PrintableString!

How about:
REDACTED(label) = ? || HEX(hash_index || HASH(serial_number || label))

...where "hash_index" is the two-octet hex representation of a single byte.

rfc6962-bis.xml
benlaurie: That's a bit sucky, nevertheless. However, I can live with it.

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+            When used, the Redacted SAN extension MUST be present in both the precertificate and the corresponding certificate.
           </t>
           <t>
-            Each INTEGER MUST have a value of zero or more. The first INTEGER indicates the total number of <spanx style="verb">?</spanx> labels in the precertificate's first DNS-ID; the second INTEGER does the same for the precertificate's second DNS-ID; etc. The last INTEGER does the same for the precertificate's zero or more CN-IDs. There MUST NOT be more INTEGERs than there are DNS-IDs (plus one, if any CN-IDs are present); if there are fewer INTEGERs than this, the shortfall is made up by implicitly repeating the last INTEGER.
+            The purpose behind hashing the redacted labels is to allow the legitimate domain owner to identify if the redacted label correlates to one of the labels they're aware of or a new, potentially unauthorized, subdomain.
```
benlaurie: s/The purpose behind hashing the redacted labels is to allow/Hashing the redacted labels allows/

rfc6962-bis.xml
AGWA: @robstradling I like encoding hash_index as a hex-encoded byte.  It will simplify the client by avoiding the need to decode an integer between the ? marks.

AGWA:
@robstradling @eranmes I think using the SKI is a great idea.  To be clear, would we use the SKI content bytes, or the extension content bytes?  In other words, would the ASN.1 OCTET STRING tag and length be included?

robstradling:
@AGWA Just the content bytes.  Not the ASN.1 OCTET STRING tag and length.

@eranmes I guess this is the current proposal then...
REDACTED(label) = ? || HEX(hash_index || HASH(subject_key_identifier || label))

Example:
Redacting "www" in "www.example.com" for https://crt.sh/?id=10557607

REDACTED(www)
  = ? || HEX(0x01 || HASH(0xA64F601E1F2DD1E7F123A02A9516E4E89AEA6E48 || 0x777777)

# echo "A64F601E1F2DD1E7F123A02A9516E4E89AEA6E48777777" | xxd -r -ps | sha256sum
4e0c8ec7c5a3c9599c91b7348175a23311c032539ac9b2633443436bb3addd3d

= ?014e0c8ec7c5a3c9599c91b7348175a23311c032539ac9b2633443436bb3addd3d.example.com

Problem: That label is 67 octets long, but the maximum length of a label is 63 octets (see RFC1034 Section 3.1).

Could we use CRC-32 instead of HASH?

AGWA:
@robstradling No, CRC-32 wouldn't stop the attack I described last month.  It would be too easy to craft a public key such that `CRC32(SKI || LABEL1) == CRC32(SKI || LABEL2)` even though `LABEL1 != LABEL2`.

Does the length matter?  These aren't actually DNS names and are never placed in the subjectAltName extension.

robstradling:
@AGWA Thanks.

I'm minded to agree with you that the label lengths don't matter.  One small niggling doubt...I expect implementers will want to reuse their SAN extension decoders to decode the Redacted SAN extension; is there any chance that the 63-octet label length is enforced at the ASN.1 decoder level?

AGWA:
@robstradling That's a good point.  We could use a more space-efficient encoding, such as base64url, but there's still a risk of exceeding 253 octets for the entire DNS name.

There's another issue: `?` is not supposed to occur in a host name; do we have to be concerned about parsers enforcing that?

robstradling:
@AGWA Instead of ?, perhaps we should see if it's possible to reserve a new ACE prefix.  Something like "rn--" (**r**edacted **n**ame) or "rl--" (**r**edacted **l**abel).

I don't know if IANA currently have a formal process for requesting such things.  This old I-D is all I could find:
https://tools.ietf.org/html/draft-ietf-idn-aceid-02

TIL that "xn--" isn't the only ACE prefix that already exists...
https://tools.ietf.org/html/rfc3490#section-5

rfc6962-bis.xml
benlaurie: Yes, I was suggesting a hex-encoded byte.

robstradling:
@AGWA Alternatively, how about we put domain names containing >=1 redacted label(s) in an OtherName field in the Redacted SAN extension?  This would completely avoid the concerns over ?s, label lengths and domain name lengths.

AGWA:
@robstradling An ACE prefix is a lot cleaner than a ? although it wouldn't solve the length problem.

The technical purity of using otherName is appealing.  I wonder if it's worth it.  It would require all implementations to write at least some new code to process.  Re-using dNSName would only require new code if the existing parser enforces length/character restrictions.

I don't have a strong opinion.  @eranmes, do you have any thoughts?

rfc6962-bis.xml

eranmes:
From what I can tell OpenSSL does not enforce the label length limitation (as it's in the DNS level, not ASN.1 encoding). OpenSSL will gladly produce CSRs with domain labels that are longer than the limit or have '?' in them (and decode it). I'll check what BoringSSL's stand on it is. Using base64, will get the labels under the length limitation but will introduce characters that are not (as far as I understand) allowed in a DNS label.

RFC5280 requires having a valid domain name for a dNSName even if it is never resolved, from
https://tools.ietf.org/html/rfc5280#section-4.2.1.6:
"When the subjectAltName extension contains a domain name system label, the domain name MUST be stored in the dNSName (an IA5String). The name MUST be in the "preferred name syntax", as specified by Section 3.5 of [RFC1034] and as modified by Section 2.1 of [RFC1123]."

However we could argue these restrictions only apply to dNSName in the subjectAltName extension and if it's in another extension then these limitations do not apply.

The options that seem available if we do want to comply:
- Base32 encoding without the padding.
- Hex encoding of a truncated hash.
- As @robstradling suggested, use otherName.

As for the ACE prefix vs ?, I notice that according to RFC5280:
"Finally, the semantics of subject alternative names that include wildcard characters (e.g., as a placeholder for a set of names) are not addressed by this specification.  Applications with specific requirements MAY use such names, but they must define the semantics.".

I'd argue our use of '?' here is OK as long as we define the semantics.

Bottom line, it seems to me the simplest course of action is hex-encode the hash index octet and the sha-256 hash, truncated to 30 bytes, so we use exactly 63 bytes per label.

eranmes:
@robstradling PTAL, per our discussion offline switched to base32 and use of the SKI.

rfc6962-bis.xml
```diff
@@ -351,18 +351,30 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
...
+            One or more labels in each <xref target="RFC6125">DNS-ID</xref> can be replaced with <spanx style="verb">REDACTED(label)</spanx> labels.
+          </t>
+          <t>
+            The redacted label is defined as follows:
+            <vspace blankLines='1' />
+            REDACTED(label) = ? || BASE32(HASH_INDEX || HASH(subject_key_identifier || label))
+            Where || denotes concatenation, HASH_INDEX is the index of a hash function in <xref target="hash_algorithms"/>, HASH is the hash function identified by HASH_INDEX, subject_key_identifier is the content bytes of the Subject Key Identifier extension (section 4.2.1.2. of <xref target="RFC5280"/>) and BASE32 is base32 encoding of the value according to <xref target="RFC4648"/>, section 6, excluding the padding.
```
benlaurie: HASH_INDEX should be a single byte (255 reserved).

Use length-encoded concatenation on general principle.

rfc6962-bis.xml
```diff
@@ -1412,21 +1424,39 @@ but it is expected there will be a variety.
...
               </t>
             </list>
           </t>
           <t>
-            A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
+            If any of these checks fail, the certificate MUST NOT be considered compliant.
           </t>
           <t>
-            If the SCT checked is for a Precertificate (where the <spanx style="verb">type</spanx> of the <spanx style="verb">TransItem</spanx> is <spanx style="verb">precert_sct_v2</spanx>), then the client SHOULD also remove embedded v1 SCTs, identified by OID 1.3.6.1.4.1.11129.2.4.2 (See Section 3.3. of <xref target="RFC6962"/>), in the process of reconstructing the TBSCertificate. That is to allow embedded v1 and v2 SCTs to co-exist in a certificate (See <xref target="v1_coexistence"/>).
+            A certificate with redacted labels where one of the redacted labels is <spanx style="verb">*</spanx> MUST NOT be considered compliant.
```
benlaurie: Would be clearer if you said something like a * in a SAN must also be a * (not redacted) in the redacted extension.

rfc6962-bis.xml
```diff
@@ -1764,7 +1794,7 @@ but it is expected there will be a variety.
       </section>
       <section title="Avoiding Overly Redacting Domain Name Labels">
         <t>
-          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. <spanx style="verb">?.com</spanx>, <spanx style="verb">?.co.uk</spanx> and other <xref target="Public.Suffix.List">public suffixes</xref>), then the domain name may be considered by clients to be overly redacted.
+          Redaction of domain name labels carries the same risks as the use of wildcards (See Section 7.2 of <xref target="RFC6125"/>, for example). If the entirety of the domain space below the unredacted part of a domain name is not controlled by a single entity (e.g. <spanx style="verb">REDACTED(label).com</spanx>, <spanx style="verb">REDACTED(label).co.uk</spanx> and other <xref target="Public.Suffix.List">public suffixes</xref>), then the domain name may be considered by clients to be overly redacted.
```
benlaurie: This phrasing is confusing - .co.uk _is_ controlled by a single entity: Nominet.

rfc6962-bis.xml
robstradling: Addressed by https://github.com/google/certificate-transparency-rfcs/pull/179/commits/f114c1fdfd04a9d1873c040adc4e86a5bcbaa4ae

rfc6962-bis.xml
robstradling: Addressed by https://github.com/google/certificate-transparency-rfcs/pull/179/commits/3de973c5cbd609666e719dac8c04caaff7207fdf

rfc6962-bis.xml
robstradling: Addressed by https://github.com/google/certificate-transparency-rfcs/pull/179/commits/b14392d5a8d3e83bf4ac04df0af66185710e01f6

PR #178, opened 2016-07-12
Explain how to ensure effective redaction

robstradling:
(and how to avoid futile redaction)

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -401,22 +401,6 @@ REDACT(label) = prefix || BASE32(index || LABELHASH(keyid || label))</artwork>
             This extension informs TLS clients of the dNSNames that were redacted and the degree of redaction, while minimizing the complexity of TBSCertificate reconstruction (as described in <xref target="reconstructing_tbscertificate"/>). Hashing the redacted labels allows the legitimate domain owner to identify whether or not each redacted label correlates to a label they know of.
           </t>
         </section>
-        <section title="Ensuring Effective Redaction">
-          <t>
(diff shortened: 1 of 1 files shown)


AGWA:
:+1:

robstradling:
@eranmes @AGWA Would it make sense to move this to the Security Considerations section?

AGWA:
@robstradling I don't like the implication that redaction is a security mechanism.  It could be a Privacy Consideration, although it also seems fine where it is currently.

rfc6962-bis.xml
```diff
@@ -401,6 +401,22 @@ REDACT(label) = prefix || BASE32(index || LABELHASH(keyid || label))</artwork>
...
+              <t>
+                Avoid registering private labels in public DNS.
+              </t>
+              <t>
+                Avoid using private labels that are predictable (e.g. <spanx style="verb">www</spanx>).
+              </t>
+            </list>
+          </t>
+          <t>
+            CAs SHOULD carefully consider each request to redact a label. When a CA believes that redacting a particular label would be futile, the CA SHOULD NOT redact it. TLS clients may have policies that forbid redaction, so redaction should only be used when it's absolutely necessary and likely to be effective.
```
eranmes: s/SHOULD/should/g, as it's not a part of the protocol, but a recommendation for CAs.

eranmes:
I agree with Rob it'd make sense to move it to another section, probably the Security Considerations section (which could be renamed Security & Privacy considerations). Since this text does not prescribe specific behaviour of a party in the protocol, it shouldn't be among the technical details of how to do redaction (it may go unnoticed there).

rfc6962-bis.xml
robstradling: Ditto for "it is RECOMMENDED that domain owners", I suppose.

robstradling:
@eranmes PTAL

eranmes:
LGTM, please squash before merging (you can now do that with the green button).

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/306817128875be6c7b232c7a85da79d418785935

PR #179, opened 2016-07-20
Address Ben's comments on simplified redaction

robstradling:


rfc6962-bis.xml
```diff
@@ -381,13 +381,13 @@ REDACT(label) = prefix || BASE32(index || LABELHASH(keyid || label))</artwork>
...
           <t>
-            <spanx style="verb">index</spanx> is the 1 byte index of a hash function in <xref target="hash_algorithms"/>.
+            <spanx style="verb">index</spanx> is the 1 byte index of a hash function in <xref target="hash_algorithms"/>. The value 255 is reserved.
           </t>
           <t>
             <spanx style="verb">LABELHASH</spanx> is the hash function identified by <spanx style="verb">index</spanx>.
           </t>
           <t>
-            <spanx style="verb">keyid</spanx> is the keyIdentifier from the Subject Key Identifier extension (section 4.2.1.2 of <xref target="RFC5280"/>), excluding the ASN.1 OCTET STRING tag and length bytes.
+            <spanx style="verb">keyid</spanx> is the keyIdentifier from the Subject Key Identifier extension (section 4.2.1.2 of <xref target="RFC5280"/>), including the ASN.1 OCTET STRING tag and length bytes.
```
eranmes: The length of the label itself is not specified. What do you think about a simple length prefix - 1 octet for specifying the length of the SKI (excluding the ASN.1 OCTET STRING tag and length bytes) and 1 octet for specifying the length of the label?

So HASH(SKI_LENGTH || SKI || LABEL_LENGTH || LABEL)
?

eranmes:
Looks good bar one comment.

eranmes:
LGTM. I'll ask Ben to have another look, but I think that for now it's safe to merge (please squash!)

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/f19b39fa361ea5fadc1cac56d49f9a2448376ba2

rfc6962-bis.xml
```diff
@@ -375,13 +375,13 @@ REDACT(label) = prefix || BASE32(index || LABELHASH(keyid || label))</artwork>
             <spanx style="verb">prefix</spanx> is the "?" character (ASCII value 63).
           </t>
           <t>
-            <spanx style="verb">||</spanx> denotes concatenation.
+            <spanx style="verb">||</spanx> denotes length-encoded concatenation. Each concatenated value is preceded by a single byte that contains the length (in bytes) of that value.
```
benlaurie: Labels don't have to be < 256 long do they?

rfc6962-bis.xml
```diff
@@ -1471,17 +1471,14 @@ but it is expected there will be a variety.
               <t>Both identities are DNS names, have the same number of labels, and each label in the subjectAltName identity has a matching label at the same position in the redactedSubjectAltName identity. Two DNS labels are matching if either:
                 <list style="symbols">
                   <t>The two labels are identical; or,</t>
-                  <t>The label from the redactedSubjectAltName identity is equal to REDACT(label from subjectAltName identity) (<xref target="redacting_labels"/>)</t>
+                  <t>Neither label is <spanx style="verb">*</spanx> and the label from the redactedSubjectAltName identity is equal to REDACT(label from subjectAltName identity) (<xref target="redacting_labels"/>).</t>
```
benlaurie: This doesn't sound right - unredacted * is fine, surely?

rfc6962-bis.xml
robstradling: https://www.ietf.org/rfc/rfc1035.txt
"Labels must be 63 characters or less"

rfc6962-bis.xml
robstradling: When * is used, it MUST be unredacted, and so both labels MUST be *.

The "The two labels are identical" case covers this.

rfc6962-bis.xml
benlaurie: Sorry, was thinking of the entire name. But where do we say the SAN must be a DNS name?

rfc6962-bis.xml
benlaurie: Ah!

benlaurie: Still a bit weird, coz the second label could never be * if they're not identical.

rfc6962-bis.xml
robstradling: "one or more non-wildcard labels **in each DNS-ID** [RFC6125] can each be replaced with a redacted label as follows"

rfc6962-bis.xml
robstradling: I couldn't find a way to eliminate that bit of weirdness without making the requirement less clear and concise.  Feel free to propose alternative wording though.

PR #180, opened 2016-07-22
The subjectAltName extension is optional.

eranmes:
CAs do not have to include the  subjectAltName extension (this is a BR). Indicate that the subjectAltName extension may be omitted even if it's to be included in the final certificate.

Also, clarify that logs MUST reject a submission where both the subjectAltName and redactedSAN extensions are present, since they make no sense.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -359,7 +359,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Redaction of Domain Name Labels" anchor="domain_name_redaction">
         <section title="Redacting Labels in Precertificates" anchor="redacting_labels">
           <t>
-            When creating a precertificate, the CA MAY omit the subjectAltName extension. Instead, the CA MUST include a <xref target="redacted_san_extension">redactedSubjectAltName</xref> extension that contains, in a redacted form, the same identities that will be included in the certificate's subjectAltName extension.
+            When creating a precertificate, the CA MAY omit the the subjectAltName extension, even if it intends to include the extension in the final certificate. Instead, the CA MUST include a <xref target="redacted_san_extension">redactedSubjectAltName</xref> extension that contains, in a redacted form, the same identities that will be included in the certificate's subjectAltName extension.
           </t>
           <figure>
             <preamble>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -359,7 +359,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
       <section title="Redaction of Domain Name Labels" anchor="domain_name_redaction">
         <section title="Redacting Labels in Precertificates" anchor="redacting_labels">
           <t>
-            When creating a precertificate, the CA MAY omit the subjectAltName extension. Instead, the CA MUST include a <xref target="redacted_san_extension">redactedSubjectAltName</xref> extension that contains, in a redacted form, the same identities that will be included in the certificate's subjectAltName extension.
+            When creating a precertificate, the CA MAY omit the the subjectAltName extension, even if it intends to include the extension in the final certificate. Instead, the CA MUST include a <xref target="redacted_san_extension">redactedSubjectAltName</xref> extension that contains, in a redacted form, the same identities that will be included in the certificate's subjectAltName extension.
```
robstradling: Why the added "the" ?

BTW, the use of "Instead" seems a bit vague.
How about: s/Instead,/If omitting the subjectAltName extension,/g

rfc6962-bis.xml
```diff
@@ -465,6 +465,9 @@ REDACT(label) = prefix || BASE32(index || LABELHASH(keyid || label))</artwork>
           Logs MUST verify that each submitted certificate or precertificate has a valid signature chain to an accepted trust anchor, using the chain of intermediate CA certificates provided by the submitter. Logs MUST accept certificates and precertificates that are fully valid according to <xref target="RFC5280">RFC 5280</xref> verification rules and are submitted with such a chain. Logs MAY accept certificates and precertificates that have expired, are not yet valid, have been revoked, or are otherwise not fully valid according to RFC 5280 verification rules in order to accommodate quirks of CA certificate-issuing software. However, logs MUST reject submissions without a valid signature chain to an accepted trust anchor. Logs MUST also reject precertificates that do not conform to the requirements in <xref target="Precertificates"/>.
         </t>
         <t>
+          Logs MUST reject submission of precertificates where both the subjectAltName and <xref target="redacted_san_extension">redactedSubjectAltName</xref> extensions are present.
```
robstradling: This added paragraph seems redundant.

The previous paragraph already says "Logs MUST also reject precertificates that do not conform to the requirements in Section 3.2", and Section 3.2 already says "the subjectAltName [RFC5280] extension MUST be omitted when the CA is using domain name redaction (Section 4.2)."

Instead of adding this paragraph, how about rewording the requirement in Section 3.2 to something programmatically enforceable?  e.g.
s/when the CA is using domain name redaction/when the redactedSubjectAltName extension is present/

eranmes: Done, adopted your suggested text for section 3.2.

rfc6962-bis.xml
eranmes: Removed the spurious 'the'.
Adopted your suggestion to replace 'Instead'

eranmes:
PTAL

PR #182, opened 2016-07-27
Be explicit about what can be redacted and how.

eranmes:
Be clear that:
- Only dNSName identities can be redacted using the first redaction mechanism.
- The second redaction mechanism (name-constrained intermediate CA cert) allows redaction of dNSName and any other field in the subjectAltName.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -399,13 +399,16 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         </section>
         <section title="redactedSubjectAltName Certificate Extension" anchor="redacted_san_extension">
           <t>
-            The redactedSubjectAltName extension is a non-critical extension (OID 1.3.101.77) that is identical in structure to the subjectAltName extension, except that dNSName identities (and only dNSName identities) MAY contain redacted labels (see <xref target="redacting_labels"/>).
+            The redactedSubjectAltName extension is a non-critical extension (OID 1.3.101.77) that is identical in structure to the subjectAltName extension, except that dNSName identities MAY contain redacted labels (see <xref target="redacting_labels"/>).
           </t>
           <t>
             When used, the redactedSubjectAltName extension MUST be present in both the precertificate and the corresponding certificate.
           </t>
           <t>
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -399,7 +399,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         </section>
         <section title="redactedSubjectAltName Certificate Extension" anchor="redacted_san_extension">
           <t>
-            The redactedSubjectAltName extension is a non-critical extension (OID 1.3.101.77) that is identical in structure to the subjectAltName extension, except that dNSName identities MAY contain redacted labels (see <xref target="redacting_labels"/>).
+            The redactedSubjectAltName extension is a non-critical extension (OID 1.3.101.77) that is identical in structure to the subjectAltName extension, except that dNSName identities (and only dNSName identities) MAY contain redacted labels (see <xref target="redacting_labels"/>).
```
robstradling: Instead of adding "(and only dNSName identities)" here, I suggest adding the following paragraph at the end of this section:

"Only DNS-IDs can be redacted using this mechanism. However, CAs can use Name Constraints (section 4.3) to allow DNS domain name labels in other types of subjectAltName entry to not appear in logs."

Conveniently, section 4.3 follows immediately afterwards.

rfc6962-bis.xml
```diff
@@ -424,6 +424,9 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
             </t>
           </list>
         </t>
+        <t>
+          Note that this mechanism allows redaction of any type of field in the subjectAltName extension.
```
robstradling: "redaction of any type of field" is misleading.  Section 4.3 is specific about the conditions under which end-entity certs may be excused from having to be logged.  If a CA doesn't meet those conditions (relating to dNSNames and IP Addresses) then they can't expect to "redact" a directoryName, SRVName, etc, even those these qualify as "any type of field".

BTW, I think we should only use the term "redaction" to refer to the section 4.2 mechanism.

If you make the change I suggested in my other comment, I think this new paragraph is probably redundant.

rfc6962-bis.xml
eranmes: Adopted your suggestion (with a small change: s/subjectAltName entry/subjectAltName entries/g)

rfc6962-bis.xml
eranmes: Removed as I've adopted your text earlier on.

PR #184, opened 2016-07-28
TLS clients may specify additional requirements for redaction.

robstradling:


rfc6962-bis.xml
```diff
@@ -357,6 +357,9 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         </t>
       </section>
       <section title="Redaction of Domain Name Labels" anchor="domain_name_redaction">
+        <t>
+          The following mechanism allows private labels to not appear in public logs, while still retaining most of the security benefits that accrue from using Certificate Transparency mechanisms. Note that TLS clients may have policies that specify additional requirements relating to the use of this redaction mechanism.
```
eranmes: Nit: Replace "may have policies that specify additional" with something more explicit indicating TLS clients have a say in whether a certificate is CT-compliant when it comes to redaction (so a certificate with over-redaction, for example, can be considered non CT-compliant even if the SCT otherwise checks out OK).
My suggested text:
"TLS clients MAY apply additional requirements relating to the use of this redaction mechanism for a certificate to be considered compliant"

robstradling: Is "MAY" appropriate here?  Currently this section only uses RFC2119 keywords for requirements on CAs.  This sentence is intended to make CAs aware that they need to consider TLS client policies, so I think "may" is appropriate.

I agree with your suggestion about explicitly mentioning CT-compliance here.

eranmes: 'may' is also fine.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/448621cc69bb8e7a25d6c73dcdfaca9abb7da14e

PR #185, opened 2016-07-28
Effective redaction requires labels to have sufficient entropy

robstradling:
Addressing James Manger's comment that "this limitation needs to be stated clearly":
https://mailarchive.ietf.org/arch/msg/trans/oC3wkUncgHP1LI3zkoerMzCB7n8


eranmes:
LGTM. 
At this point I feel like we're adding disclaimers so that nobody can later say the disadvantages of redaction weren't clear in hindsight.
*sigh*

PR #186, opened 2016-07-29
Ticket 111: Use cached_info to indicate prior knowledge of CT compliance

robstradling:


eranmes:
To capture our out-of-band discussion:
I don't see an issue with the text as-is.
For simplicity, though, I'd prefer to have only a single value in the CachedObject that the client could use to indicate  it is not interested in the server sending it any TransItems at all.

It is not guaranteed that, should a client perform an inclusion check, that the inclusion status will ever be fed back to the certificate status (as held by the client), so the 2nd value specified may be moot. (that is from trying to implement inclusion checking in Chrome).

My suggestion, then, is to only have a single value where the client indicates it is not interested in any TransItems and let the client decide what that means for it.

robstradling:
OK, I'll make that change.

robstradling:
Actually, "not interested in the server sending it any TransItems" is too broad.  The transparency_info TLS extension could concievably be used to send TransItems that are unrelated to the cached cert.  e.g. for gossip.
So I'll just mention the TransItems for SCTs and inclusion proofs.

robstradling:
@eranmes PTAL

rfc6962-bis.xml
```diff
@@ -1363,6 +1363,14 @@ entry specified by <spanx style="verb">start</spanx>.
...
         </t>
       </section>
+      <section title="cached_info TLS Extension">
+        <t>
+          Provided that a TLS client includes the <spanx style="verb">cached_info</spanx> (<xref target="RFC7924"/>) extension type in the ClientHello with a <spanx style="verb">CachedObject</spanx> of type <spanx style="verb">ct_compliant</spanx> (see <xref target="tls_cachedinfo_extension"/>) and at least one <spanx style="verb">CachedObject</spanx> of type <spanx style="verb">cert</spanx>, and provided that the TLS server then sends a modified Certificate message (as described in section 4.1 of <xref target="RFC7924"/>) and includes the <spanx style="verb">transparency_info</spanx> extension in the ServerHello, the TLS server SHOULD NOT include any <spanx style="verb">TransItem</spanx> structures of type <spanx style="verb">x509_sct_with_proof_v2</spanx>, <spanx style="verb">x509_sct_v2</spanx>, <spanx style="verb">precert_sct_with_proof_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> in the <spanx style="verb">TransItemList</spanx> that it includes in the <spanx style="verb">transparency_info</spanx> extension.
```
eranmes: To be clear, does the TLS client have to also send the transparency_info extension  for this to have effect?

eranmes:
LGTM with one question, once that's clarified, can be merged.

rfc6962-bis.xml
robstradling: Yes.  For clarity, let's state that explicitly.

robstradling:
@eranmes PTAL

rfc6962-bis.xml
```diff
@@ -1363,6 +1363,14 @@ entry specified by <spanx style="verb">start</spanx>.
...
+        <t>
+          Provided that a TLS client includes the <spanx style="verb">transparency_info</spanx> extension type in the ClientHello and includes the <spanx style="verb">cached_info</spanx> (<xref target="RFC7924"/>) extension type in the ClientHello with a <spanx style="verb">CachedObject</spanx> of type <spanx style="verb">ct_compliant</spanx> (see <xref target="tls_cachedinfo_extension"/>) and at least one <spanx style="verb">CachedObject</spanx> of type <spanx style="verb">cert</spanx>, and provided that the TLS server then sends a modified Certificate message (as described in section 4.1 of <xref target="RFC7924"/>) and includes the <spanx style="verb">transparency_info</spanx> extension in the ServerHello, the TLS server SHOULD NOT include any <spanx style="verb">TransItem</spanx> structures of type <spanx style="verb">x509_sct_with_proof_v2</spanx>, <spanx style="verb">x509_sct_v2</spanx>, <spanx style="verb">precert_sct_with_proof_v2</spanx> or <spanx style="verb">precert_sct_v2</spanx> in the <spanx style="verb">TransItemList</spanx> that it includes in the <spanx style="verb">transparency_info</spanx> extension.
```
eranmes: While correct, I personally find this sentence a bit long and hard to parse (i.e. have to load all the conditions into memory before figuring out what they apply to). How about reversing it to something like:
The TLS server SHOULD NOT include any ... if all the following are met:
- TLS client includes transparency_info AND cached_info extension in the handshake
- ...

eranmes:
Still LGTM with a minor suggestion for the first changed paragraph.

eranmes:
Looks much better, thanks for splitting it up.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/3cf2e309f69a6995bcdc7c42caf4b3818b317f67

PR #187, opened 2016-08-01
Allowing more space for nodes in proofs.

eranmes:
The way it's currently specified there can only be 7 repeated nodes for SHA-256 hash, for example (which fit in 255 bytes), which is not enough for large logs.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -744,7 +744,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         LogID log_id;
         uint64 tree_size_1;
         uint64 tree_size_2;
-        NodeHash consistency_path&lt;1..2^8-1&gt;;
+        NodeHash consistency_path&lt;1..2^16-1&gt;;
     } ConsistencyProofDataV2;</artwork>
         </figure>
         <t>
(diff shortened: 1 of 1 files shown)


eranmes:
Rob/Ben: PTAL - Unless I'm missing anything, the way ConsistencyProofDataV2 / InclusionProofDataV2 are specified there's a vector of NodeHash but the total size of this vector may be up to 255 bytes. Given a SHA-256 hash is 32 bytes, that'd only allow for 7 nodes.

robstradling:
I was just about to write that it's a vector of up to 255 `NodeHash`es rather than bytes...but I just re-read RFC5246 section 4.3 and found I was wrong.  For fixed-length vectors, "the length declares the number of bytes, not the number of elements".  Judging from the examples provided (in 5246 section 4.3), the same is true for variable-length vectors.

Note that 5246 section 4.3 also says:
  "The length of an encoded vector must be an even multiple of the length of a single element (for example, a 17-byte vector of uint16 would be illegal)."

So I think that "1..2^8-1" and "1..2^16-1" are both illegal and that we need to express the maximum length as "an even multiple of the length of" HASH_SIZE !!

eranmes:
This is going to lead to an interesting design constraint on all future HASH_SIZEs.

robstradling:
Maybe we should get rid of the "NodeHash" definition and go back to using "opaque" instead, just so we can avoid this "interesting design constraint" ?

eranmes:
Closing this in favour of https://github.com/google/certificate-transparency-rfcs/pull/189.

PR #189, opened 2016-08-17
Ticket 158: Make NodeHash variable-length

eranmes:
See https://trac.tools.ietf.org/wg/trans/trac/ticket/158 on why it's inconvenient to have fixed-length fields specified in the various data structures when the length depends on the log the data is parsed from.

This PR also allows more space for nodes in proofs, covering PR #187.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -521,7 +521,7 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
 
     struct {
         uint64 timestamp;
-        opaque issuer_key_hash[HASH_SIZE];
+        opaque issuer_key_hash&lt;32..2^8-1&gt;;
         TBSCertificate tbs_certificate;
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -612,6 +612,9 @@ d0 d1   d2 d3           d0 d1   d2 d3  d4 d5</artwork>
         <t>
           <spanx style="verb">sth_extensions</spanx> matches the STH extensions of the corresponding STH.
         </t>
+        <t>
+          The length of the NodeHash MUST match HASH_SIZE of the log.
```
robstradling: "the NodeHash" doesn't sound quite right, because "the" implies it's singular.

I suggest dropping the "the".  So just "The length of NodeHash..."

robstradling:
LGTM with one nit (and hoping a max length of 255 bytes is long enough in the future!)

rfc6962-bis.xml

eranmes:
Thanks for the quick review Rob, PTAL.

PR #193, opened 2016-08-17
TLS clients may only evaluate compliance if they send both the transparency_info and status_request TLS extensions

robstradling:


eranmes:
I overall agree with the sentiment expressed, just trying to find a more concise/consistent way to express it. It seems we have a very similar/related text in https://tools.ietf.org/html/draft-ietf-trans-rfc6962-bis-18#section-10.2.1:
"TLS clients that support the "transparency_info" TLS extension SHOULD include it in ClientHello messages, with empty "extension_data"."

How about changing the "SHOULD" to "MUST" and append  "in order to be able to evaluate CT compliance" or something of that sort?

robstradling:
That existing text only covers the "transparency_info" TLS extension.  I don't see any other existing text that says that TLS clients MUST send the "status_request" TLS extension.

So, if I were to instead change that "SHOULD" to a "MUST", I'd have to add another sentence saying that sending "status_request" is also a MUST.  That doesn't seem any more concise than the text I've proposed.

Also, the text I've proposed is intended to make sense in cases where the TLS client supports CT but doesn't enforce it 100% of the time.  For example, it might be useful for a TLS client to have (1) a debug setting where the sending of certain TLS extensions is suppressed, or (2) a blacklist of hostnames to which certain TLS extensions should not be sent due to TLS server bugs...but I don't think your proposal of changing that "SHOULD" to "MUST" would allow such things.

Would you be happy with my proposed text?

eranmes:
Good point about the SHOULD/MUST change. I agree we should not change to MUST in that case.

eranmes:
Re-reading the text with this context, LGTM.

robstradling:
Merged at https://github.com/google/certificate-transparency-rfcs/commit/a58dc00dd8bdfc30c639f48880d986d0f73c770a

PR #194, opened 2016-08-17
Rename get-entry-for-certificate to get-entry-for-tbscertificate.

robstradling:
Clarify which TBSCertificate to hash.

### DIFF
--- a/rfc6962-bis.xml
+++ b/rfc6962-bis.xml
@@ -1176,16 +1176,16 @@ entry specified by <spanx style="verb">start</spanx>.
 	</t>
       </section>
 
-      <section title="Get Entry Numbers for Certificate">
+      <section title="Get Entry Numbers for TBSCertificate">
 	<t>
-	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-certificate
+	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-tbscertificate
 	</t>
 	<t>
           <list style="hanging">
             <t hangText="Inputs:">
               <list style="hanging">
                 <t hangText="hash:">
-                  A base64 encoded HASH of a <spanx style="verb">TBSCertificate</spanx>.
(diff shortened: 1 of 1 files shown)


rfc6962-bis.xml
```diff
@@ -1176,16 +1176,16 @@ entry specified by <spanx style="verb">start</spanx>.
...
-	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-certificate
+	  GET https://&lt;log server&gt;/ct/v2/get-entry-for-tbscertificate
 	</t>
 	<t>
           <list style="hanging">
             <t hangText="Inputs:">
               <list style="hanging">
                 <t hangText="hash:">
-                  A base64 encoded HASH of a <spanx style="verb">TBSCertificate</spanx>.
+                  A base64 encoded HASH of a <spanx style="verb">TBSCertificate</spanx> for which the log has previously issued an SCT. (Note that a precertificate's TBSCertificate can be reconstructed from the corresponding certificate as described in <xref target="reconstructing_tbscertificate"/>).
```
eranmes: s/can be reconstructed/MUST be reconstructed ?
Otherwise it won't work...

eranmes:
LGTM otherwise.

rfc6962-bis.xml
robstradling: I copied that sentence verbatim from elsewhere in the document.

IIRC (but please do tell me if I'm making this up ;-) ), @benlaurie wanted us to only use MUST, SHOULD, etc, when talking about actors in the CT ecosystem (e.g. "TLS clients MUST...").  Here, we're talking about an object (a precertificate), not an actor.

Perhaps we could change "can be" to "is"?

eranmes: s/can be/is/ also SGTM.
You could re-phrase to say something like 'callers MUST reconstruct the TBSCertificate from the corresponding certificate for precertificate entries, as described in ...'.
But I don't mind either - as long as it's clear a client should not blindly hash the TBSCertificate for use in this call.

robstradling:
Merged at https://github.com/robstradling/certificate-transparency-rfcs/commit/2e4904d8672a7d6ea639d0a625f8a02e6b948cf2

PR #196, opened 2016-08-26
Convert 6962-bis to kramdown.

robstradling:


eranmes:
I was thinking we'd leave 6962-bis.xml around and rename it to OBSOLETE-6962-bis.xml, so git blame would be easy to use?

AGWA:
@eranmes You can specify the starting revision as an optional argument to `git blame`, so instead of leaving the XML file in the repository, you could tag the last XML revision as `pre-kramdown` and then use `git blame` like this:

```
git blame pre-kramdown 6962-bis.xml
```

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -0,0 +1,2193 @@
...
+is defined as follows:
+
+The proof for the single leaf in a tree with a one-element input list D\[1] =
+{d(0)} is empty:
+
+~~~~~~~~~~~
+PATH(0, {d(0)}) = {}
+~~~~~~~~~~~
+
+For n > 1, let k be the largest power of two smaller than n.  The proof for the
```
eranmes: A spurious whitespace in this life ('  The proof...') which may have existed in the original.

robstradling: Hmmm, it looks like it wasn't there in the original.  I guess my converter must've converted all occurrences of `". "` to `".  "`

eranmes:
LGTM - I've only thoroughly reviewed the .txt files to verify there are no differences in content.
HTMLs look the same.

I note we've gained a new 'Authors' addresses' section, which we could probably do without, and a few spurious double spaces, which we can clean up later.

robstradling:
The Author's Addresses section is added by xml2rfc.  It was there before the kramdown conversion.

robstradling:
@eranmes I've just added a `pre-kramdown` tag, as @AGWA suggested.

PR #200, opened 2016-10-28
Ticket 160: Add get-sths.

robstradling:


robstradling:
@phad This covers your proposal for a get-sths-between-times API.

It doesn't address your get-sth-at-time proposal.  (I'm not sure if merging this with get-sth, with an optional input parameter, would make sense or not, if we're saying that it's optional for logs to retain historical STHs).

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1402,6 +1402,45 @@ Note that it is possible for a certificate to be logged more than once. If that
 is the case, the log MAY return more than one entry index. If the certificate is
 present in the log, then the log MUST return at least one entry index.
 
+## Retrieve Signed Tree Heads between Two Times
+
+GET https://\<log server>/ct/v2/get-sths
```
phad: Are you going to say something about this end-point being optional?
Or were you planning to leave that until the decision that it should be optional is made?

phad:
Thanks Rob.
I agree with your caution over merging the 'get-sth-at-time' request with the existing get-sth if support for the whole historical STHs feature is optional.  I think it would lack clarity to say that 'get-sth' is a required end-point while specifying a parameter that implementations can omit support for.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1402,6 +1402,45 @@ Note that it is possible for a certificate to be logged more than once. If that
...
+  : a later NTP Time [RFC5905], measured in milliseconds since the epoch
+    (January 1, 1970, 00:00 UTC), ignoring leap seconds.
+
+Outputs:
+
+: sths:
+  : an array of base64 encoded `TransItem` structures of type
+    `signed_tree_head_v2`, signed by this log.
+
+The `start` and `end` parameters SHOULD be within the range 0 <= x < `timestamp`
```
robstradling: If a parameter is optional, it's explicitly marked as optional.  e.g., see the description of the "second" input to "get-sth-consistency".

draft-ietf-trans-rfc6962-bis.md
robstradling: It already says "7. Optional Client Messages"

draft-ietf-trans-rfc6962-bis.md
phad: Got it, thanks.

draft-ietf-trans-rfc6962-bis.md
phad: Yeah I was about to say, the preface for 'optional messages' says they are 'not required for correct operation of logs or their clients', but you might make the case that get-sths is pretty important for correct operation of an auditing monitor...

PR #201, opened 2016-10-28
Ticket 161: Add 'try later' error code for add-chain and add-pre-chain.

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1065,6 +1065,7 @@ Error codes:
 | unknown anchor  | The last certificate in the chain both is not, and is not certified by, an accepted trust anchor. |
 | bad chain       | The alleged chain is not actually a chain of certificates.                                        |
 | bad certificate | One or more certificates in the chain are not valid (e.g., not properly encoded).                 |
+| try later       | The log is operational but is temporarily unable to accept new submissions.                       |
```
sleevi: Does this response overlap with 503 - Service Unavailable? It seems like using that (plus a Retry-After header) could indicate this as well.

https://tools.ietf.org/html/rfc7231#section-6.6.4

jsha: 503 + Retry-After is already specified: https://github.com/google/certificate-transparency-rfcs/blob/master/draft-ietf-trans-rfc6962-bis-19.txt#L1281-L1294

And it is already implemented in at least one CT client: https://github.com/google/certificate-transparency/blob/master/go/client/logclient.go#L259-L269 (this is the one Let's Encrypt uses).

So I think there is no change needed here?

robstradling: @sleevi @jsha Good point.  I'll abandon this PR.

PR #203, opened 2016-11-17
Adding 1st draft of monitoring api.

eranmes:
This is a very, very rough 1st draft of the monitoring API.
I've created it so that I have somewhere to put the methods I'd like to remove from 6962-bis.

PR #204, opened 2016-11-17
Remove optional API.

eranmes:
The optional methods removed in this PR have been moved to the monitoring API in PR #203.


eranmes:
Rob, Ben, PTAL.
Note this is a no-op change: Logs could still implement these methods in the same namespace as the log itself (i.e. under /ct/v2/), so this is just shuffling text around.

PR #205, opened 2016-11-24
Updating 'how to push i-d' instructions.

eranmes:
The instructions are different since the move to markdown.
Note that compared to the XML file, where the *next* version number was always kept, the markdown version has the *current* version number and so has to be bumped first, before uploading an i-d.


eranmes:
@robstradling PTAL.

PR #212, opened 2016-12-12
Major changes update

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -227,6 +227,10 @@ community. The major changes are:
 - Logs IDs: each log is now identified by an OID rather than by the hash of its
   public key.
 
+- TransItem structure: this new data structure, which encapsulates most types of
+  CT data, can be used anywhere that SignedCertificateTimestampList was used in
```
eranmes: That's not quite accurate: A TransItemList can now be used everywhere a SignedCertificateTimestampList could be used in RFC6962.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -235,9 +235,9 @@ community. The major changes are:
   TransItem structure, which eases extensibility and simplifies the leaf
   structure by removing one layer of abstraction.
 
-- Signatures in SCTs for X.509 Certificates no longer cover the entire
-  certificate. Instead, leaf entries for Precertificates and X.509
-  certificates include the same data.
+- Unified leaf format and SCT signatures: the structure for both certificate and
```
eranmes: Drop " and SCT signatures" ? It is not immediately obvious how the SCT signatures relate, maybe best to not mention that entirely and let the reader look up the unified leaf format definition.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -250,6 +250,27 @@ community. The major changes are:
...
+- CT TLS extension: the `signed_certificate_timestamp` TLS extension has been
+  replaced by the `transparency_info` TLS extension.
+
+- Other TLS extensions: `status_request_v2` may be used (in the same
+  manner as `status_request`); `cached_info` may be used to avoid sending the
+  same complete SCTs and inclusion proofs to the same TLS clients multiple
+  times.
+
+- TLS Feature extension: this certificate extension may be used by a CA to
+  indicate that CT compliance is required.
```
eranmes: A bit of a minefield as some things in this document say when a certificate isn't "CT compliant"  but UAs can require more than just one SCT, for example, to consider a certificate to be "CT compliant".
Not sure what to suggest here, though.

draft-ietf-trans-rfc6962-bis.md
robstradling: Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/34e488f5b93adee9986c95db5cc8685ba9b5d4a9

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -250,6 +250,27 @@ community. The major changes are:
...
+
+- Verification algorithms: added detailed algorithms for verifying inclusion
+  proofs, for verifying consistency between two STHs, and for verifying a root
+  hash given a complete list of the relevant leaf input entries.
+
+- Extensive clarifications and editorial work.
```
eranmes: nit: That's evident from 'diff'-ing the two documents, I suggest we drop it? (I agree that's a major change - but with the view of providing highlight of technical changes, that one seems unrelated).

eranmes:
LGTM once the comments are addressed.

draft-ietf-trans-rfc6962-bis.md
robstradling: I think "may be used by a CA to indicate that CT compliance is required" is entirely consistent with what we specify later in the document.

draft-ietf-trans-rfc6962-bis.md
robstradling: The equivalent "major differences" section in RFC5246 (TLS 1.2) ends with...
  "  - The usual clarifications and editorial work."
...so I see no reason why we can't keep this in.

PR #220, opened 2016-12-28
Add name constrained option (to the Redaction I-D)

robstradling:


redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
+an intermediate CA certificate or intermediate CA precertificate (to which the
+certificate identified by the `certID` of that `SingleResponse` chains) that
+meets the requirements in {{name_constrained}}. This extends section TBD of CT
+v2 [I-D.ietf-trans-rfc6962-bis], which specifies that each SCT always
+corresponds to the certificate identified by the `certID` of that
+`SingleResponse` or to a precertificate that corresponds to that certificate.
+
+Each SCT (and optional corresponding inclusion proof and STH) included by a
+certification authority in a Transparency Information X.509v3 extension in a
+certificate MAY correspond to an intermediate CA certificate or intermediate CA
```
eranmes: This paragraph and the previous two paragraphs share a lot in common - the only difference being the SCT delivery method and, in the case of SCTs from the OCSP response, the specification of the certID the SCT refers to.

Can we refactor the text so that we don't repeat most of it 3 times?

redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
+            }
+          }
+        }
+      }
+~~~~~~~~~~~
+
+### Presenting SCTs, Inclusion Proofs and STHs
+
+Each SCT (and optional corresponding inclusion proof and STH) presented by a TLS
+server MAY correspond to an intermediate CA certificate or intermediate CA
```
eranmes: Should we limit the number of name-constained CA certs in the chain to one?

That would effectively allow us to say if there's such a name-constrained CA cert (that meets the requirements in {{name_constrained}}), then it should be only one, and all SCTs observed in the connection (from other certs in the chain, TLS handshake or stapled OCSP response) refer to it.

That would extend the section in 6962-bis about SCT corresponding to the server certificate in a  deterministic  and less confusing way.

redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
+precertificate (to which the certificate chains) that meets the requirements in
+{{name_constrained}}. This extends section TBD of CT v2
+[I-D.ietf-trans-rfc6962-bis], which specifies that each SCT always corresponds
+to a precertificate that corresponds to that certificate.
+
+### Matching an SCT to the Correct Certificate
+
+Before considering any SCT to be invalid, a TLS client MUST attempt to validate
+it against the server certificate and against each of the zero or more suitable
+name-constrained intermediates in the chain. These certificates may be evaluated
```
eranmes: See my comment above - I suggest we limit it to only one name-constrained intermediate CA certificate and the client checks SCTs either against the server certificate or that intermediate.

redaction/draft-strad-trans-redaction.md
```diff
@@ -187,14 +300,14 @@ in the certificate, TLS clients MUST also:
 
 # Security Considerations
 
-## Avoiding Overly Redacting Domain Name Labels
+## Avoiding Overly Redacted Domain Names
```
eranmes: Please add a TODO(eranm): here to describe harm to the CT ecosystem that would arise from the use of redaction being too widespread.

redaction/draft-strad-trans-redaction.md
```diff
@@ -88,6 +94,64 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC2119].
 
+# Wildcard Certificates
+
+A certificate containing a DNS-ID [RFC6125] of `*.example.com` could be used to
+secure the domain `topsecret.example.com`, without revealing the string
```
robstradling: Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/10cd9abd14675c02d71ff0d0e399b143d7f0dd73

(FWIW, I think either "string" or "label" would be fine here.  The string is indeed a label, but if it had no significance beyond being just a label there wouldn't be any need to keep it secret!)

redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
+To ensure effective redaction, CAs and domain owners should note the privacy
+considerations ({{privacy_considerations}}).
+
+TODO(eranm): Do we need to further expand (either here or in the following
+subsections) on when each of the mechanisms is/isn't suitable?
+
+## Using Wildcard Certificates    {#wildcard_certificates}
+
+A certificate containing a DNS-ID [RFC6125] of `*.example.com` could be used to
+secure the domain `topsecret.example.com`, without revealing the string
```
robstradling: (re-confirming, in case my confirmation got lost ;-) )

Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/10cd9abd14675c02d71ff0d0e399b143d7f0dd73

(FWIW, I think either "string" or "label" would be fine here. The string is indeed a label, but if it had no significance beyond being just a label there wouldn't be any need to keep it secret!)

redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
 # Requirements Language
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in [RFC2119].
 
-# Redacting Labels in Precertificates    {#redacting_labels}
+# Redaction Mechanisms
+
+We propose three mechanisms to deal with this conundrum, in increasing order of
```
robstradling: Good point.  Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/8ffb8cf7b7bc0cd94867f15fffb96d6f4d1b4c2f (by repeating/summarizing the first half of the Abstract).

redaction/draft-strad-trans-redaction.md
```diff
@@ -69,26 +75,133 @@ certificates.
...
+  end-entity certificate ({{name_constrained}}) covers more, but not all, use
+  cases.
+
+* Therefore, we define a domain label redaction mechanism ({{redacting_labels}})
+  that covers all use cases, at the cost of considerably increased
+  implementation complexity.
+
+We anticipate that TLS clients may develop policies that impose additional
+compliancy requirements on the use of the {{name_constrained}} and
+{{redacting_labels}} mechanisms.
```
robstradling: I don't think there are any restrictions on wildcard certs worth mentioning here.

AFAIK, there is no absolute "cannot have a wildcard for a domain in the PSL" rule at the moment.  The BRs do flag this as something for CAs to watch out for, but they also say:
"A CA is not prohibited from issuing a Wildcard Certificate to the Registrant of an entire gTLD, provided that control of the entire namespace is demonstrated in an appropriate way." (BRs v1.4.1 section 3.2.2.6)

The EV Guidelines prohibit Wildcard EV certs, but we already mention this in the "Using Wildcard Certificates" subsection.  (Also, this is an industry guideline rule rather than a TLS client policy rule).

redaction/draft-strad-trans-redaction.md
robstradling: The purpose of this PR is to add the text about the name-constrained intermediate redaction mechanism that we removed from 6962-bis, and to then do whatever editorial work is necessary to ensure the document still roughly makes sense.

Limiting the number of name-constrained intermediate certs in the chain to one may well make sense, but since it'd be a change to the mechanism that used to be in 6962-bis, let's tackle it in a separate PR.  Is that OK?

redaction/draft-strad-trans-redaction.md
robstradling: I agree that this repetition seems sub-optimal, but I wanted to make sure I captured the requirements precisely.  Writing it verbosely seemed like the safest way to accomplish this, even if we subsequently refactor the text.  (If I'd refactored at the same time, I fear I might've inadvertently lost some of the subtleties).

TBH, it's not immediately obvious to me how to refactor this section, so do you mind if we punt it to a separate PR?

redaction/draft-strad-trans-redaction.md
robstradling: Let's do this is a separate PR.

redaction/draft-strad-trans-redaction.md
robstradling: Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/13a03a913232e4072d171eac42a4b6ce95b87e40

eranmes:
LGTM

PR #225, opened 2017-03-29
Issue #172 - Pick one structure for multiple TransItems

bifurcation:


robstradling:
There are some remnants of the "_sct_with_proof_" machinery still in the document.  I'll do a follow-up PR to address this.

PR #226, opened 2017-03-29
Issue #166 - State the log parameters in the section that defines a log

bifurcation:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -555,6 +555,51 @@ new entries to its Merkle Tree and sign the root of the tree.
 Log operators MUST NOT impose any conditions on retrieving or sharing data from
 the log.
 
+## Log Parameters
```
eranmes: There are still a few sections which refer to the #metadata section, the section should either be given a new reference or have the '#metadata' reference added.

PR #228, opened 2017-03-29
Issue #182 - Clarify notation in the Merkle tree section

bifurcation:


eranmes:
Overall LGTM, would like a second opinion since I recall Rob looking into this in the past.

robstradling:
@eranmes Please ask Ben, Emilia or Adam for a second opinion.  The Merkle Tree stuff is the part of CT I feel least qualified to offer an opinion on.  :-)

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -265,8 +265,8 @@ algorithms, see {{hash_algorithms}}. The hashing algorithm in use is referred to
 as HASH throughout this document and the size of its output in bytes as
 HASH_SIZE. The input to the Merkle Tree Hash is a list of data entries; these
 entries will be hashed to form the leaves of the Merkle Hash Tree. The output is
-a single HASH_SIZE Merkle Tree Hash. Given an ordered list of n inputs, D\[n] =
-{d(0), d(1), ..., d(n-1)}, the Merkle Tree Hash (MTH) is thus defined as
+a single HASH_SIZE Merkle Tree Hash. Given an ordered list of n inputs, D^(n) =
```
benlaurie: If we're going to go down this path, which I do quite like, I think it'd be a little less weird to use a subscript and no brackets ... i.e. D_n.

Not everyone can read TeX, either, so might be worth explaining the notation. :-)

eranmes: Do I understand correctly that your only suggested change regarding this PR is to change D^(n) to D_n (throughout) ?
If so, seems reasonable to me.

benlaurie: Mostly. We struggled with getting the notation for this right from the beginning. Not convinced this is less confusing, but I don't think it is _more_ confusing.

The one other change I might make I will suggest in a moment. :-)

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -312,57 +312,57 @@ list of missing nodes required to compute the nodes leading from a leaf to the
...
+PATH(m, D^(n)) = PATH(m, D[0:k]) : MTH(D[k:n]) for m < k; and
 
-PATH(m, D[n]) = PATH(m - k, D[k:n]) : MTH(D[0:k]) for m >= k,
+PATH(m, D^(n)) = PATH(m - k, D[k:n]) : MTH(D[0:k]) for m >= k,
 ~~~~~~~~~~~
 
-where : is concatenation of lists and D\[k1:k2] denotes the length (k2 - k1)
-list {d(k1), d(k1+1),..., d(k2-1)} as before.
+Here, : is a concatenation of lists, and as before, D\[k1:k2] denotes the list
+{d[k1], d[k1+1], ..., d[k2-1]} of length (k2 - k1).
```
benlaurie: This is the core problem ... D[k1:k2] is really a new tree, so having different notation and numbering than the original tree is confusing. It occurs to me that we could say something like:

D[k1:k2] = D'_{k2 - k1} denotes the list {d'[0] = d[k1], d'[1] = d[k1 + 1], ... , d'[k2 - k1 - 1] = d[k2 - 1]}

and that fixes the weirdness of the list numbering...

eranmes:
Closing this in favour of https://github.com/google/certificate-transparency-rfcs/pull/251, simply because merging Richard's original PR with my and Ben's changes became too messy.

PR #230, opened 2017-03-29
Issue #171 - Change LogIDs from OID to uint32

bifurcation:


eranmes:
Closing per https://trac.ietf.org/trac/trans/ticket/171#comment:4.
If you feel we need to add an explanation about the reasoning for the use of OIDs in the document, do let me know.

PR #231, opened 2017-03-29
Issue #187 - Used IANA-managed OIDs

bifurcation:


eranmes:
Closing since, per @robstradling (https://trac.ietf.org/trac/trans/ticket/187#comment:3), this OID is used in other places and there's work going on to let somebody else administer it.

PR #237, opened 2017-04-03
Issue #169 - Don't guess at STHs

bifurcation:


eranmes:
I suggest leaving the STH as an output parameter for get-all-by-hash which is a new method in 6962-bis, as it would allow us to implement ekr's suggestion.

Otherwise looks good.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1122,10 +1122,6 @@ Outputs:
     `tree_size_1` MUST match the `first` input. If the `sth` output is omitted,
     then `tree_size_2` MUST match the `second` input.
 
-  sth:
-  : A base64 encoded `TransItem` of type `signed_tree_head_v2`, signed by this
-    log.
-
```
robstradling: Please also update the paragraph just above the "Outputs:", which still refers to this deleted "sth" output parameter.  It currently says:
```Both tree sizes must be from existing v2 STHs.  However, because of skew, the receiving front-end may not know one or both of the existing STHs.  If both are known, then only the "consistency" output is returned.  If the first is known but the second is not (or has been omitted), then the latest known STH is returned, along with a consistency proof between the first STH and the latest.  If neither are known, then the latest known STH is returned without a consistency proof.```

File: draft-ietf-trans-rfc6962-bis.md [hunk 2]
```diff
@@ -1166,12 +1163,8 @@ Outputs:
     `inclusion_path` array of Merkle Tree nodes proves the inclusion of the
     chosen certificate in the selected STH.
 
-  sth:
-  : A base64 encoded `TransItem` of type `signed_tree_head_v2`, signed by this
-    log.
-
 > Note that no signature is required for the `inclusion` output as it is used to
-> verify inclusion in the selected STH, which is signed.
+> verify inclusion under the selected STH, which is signed.
```
robstradling: Let's be consistent.  "in" or "under", but not a mixture.  I see an occurrence of "inclusion...in the selected STH" just a few lines above.

eranmes:
(adding the question label to indicate it's pending decision/discussion on trans)

eranmes:
Closing per https://trac.ietf.org/trac/trans/ticket/169#comment:4, since it seems quite a few people found this feature useful.

PR #239, opened 2017-04-03
Issue #185 - Don't violate BCP 190

bifurcation:


eranmes:
LGTM - I'm OK with that, equally happy with Mark Nottingham's suggestion of removing the 'ct/v2' part of the URI

robstradling:
I'd drop the "This protocol can be updated to allow more flexibility if it is needed" statement, I think.  IIRC, we've already removed several statements (that were present in RFC6962) along the lines of "We could maybe change X to Y in future if it's needed".

Other than that, LGTM.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -959,6 +959,13 @@ using the "application/x-www-form-urlencoded" format described in the "HTML 4.01
 Specification" [HTML401]. Binary data is base64 encoded [RFC4648] as specified
 in the individual messages.
 
+Clients are configured with a base URL for a log and construct URLs for
+requests by appending suffixes to this base URL.  This structure places some
+degree of restriction on how log operators can deploy these services, as noted
+in {{?BCP190}}.  However, operational experience with version 1 of this
```
eranmes: You could refer to RFC7320, which seems to be the same?

eranmes:
Merged in PR #249  instead.

PR #183, opened 2016-07-27
Use DNS-ID consistently

robstradling:

PR #240, opened 2017-04-03
Issue #183 - Don't violate the TLS Feature extension definition

bifurcation:


robstradling:
@eranmes @bifurcation
This commit only removed about half of the document's references to TLS Feature.  I'm guessing that the intent was to remove the other half too?

I'm mildly concerned that saying absolutely nothing about TLS Feature leaves things a bit underspecified.  Should we add a statement that says something like "Using the TLS Feature extension to mandate the use of the transparency_info TLS extension is NOT RECOMMENDED." ?

PR #241, opened 2017-04-03
Issue #184 - Remove unnecessary restrictions on clients

bifurcation:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1626,12 +1626,22 @@ client uses `get-all-by-hash`, then it will already have the new STH.
...
+
+It is up to a client's local policy which forms of evidence are sufficient for
+compliance.  For example, a client might require certificates to come with some
+minimum number of SCTs from among a collection of trusted logs, or require a
+single inclusion proof to an audited tree head.
+
+Servers may provide evidence of logging via any of the channels described in
+{{tls_servers}}.  In order to ensure that any information sent by the server is
+successfully received by the client, clients that intend to evaluate compliance
+MUST support both the `transparency_info` and `status_request` TLS extensions.
```
eranmes: I think the document says something similar somewhere else, I'll check.

eranmes:
Overall looks good, adding Rob for 2nd opinion.

draft-ietf-trans-rfc6962-bis.md
eranmes: Yes, there's text in section 8.2.1. about clients having to support the three SCT delivery mechanisms - How about moving the last sentence into that section? 
Section 8.2.1. should explicitly say that TLS clients MUST support both the `transparency_info` and `status_request` TLS extensions for properly receiving SCTs, but it seems to me we should not force TLS clients to fully support processing the `status_request` / `status_request_v2` extensions (i.e. they could use it just for extracting SCTs).

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1572,13 +1572,20 @@ elements.
...
-implement all of the three mechanisms by which TLS servers may present SCTs (see
-{{tls_servers}}). TLS clients MAY also accept SCTs via the `status_request_v2`
-extension ([RFC6961]). TLS clients that support the `transparency_info` TLS
-extension SHOULD include it in ClientHello messages, with empty
-`extension_data`. TLS clients may also receive inclusion proofs in addition to
-SCTs, which should be checked once the SCTs are validated.
+Servers may provide evidence of logging via any of the channels described in
+{{tls_servers}}.  In order to ensure that any information sent by the server is
+successfully received by the client, clients that intend to evaluate compliance
+MUST support both the `transparency_info` and `status_request` TLS extensions.
```
robstradling: Why add this paragraph?  The first sentence is irrelevant to this section.  The second sentence is already covered by the next paragraph.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1572,13 +1572,20 @@ elements.
...
+
+TLS clients receive SCTs alongside or in certificates. In order to ensure that
+any information sent by the server is successfully received by the client, TLS
+clients MUST implement all of the three mechanisms by which TLS servers may
+present SCTs (see {{tls_servers}}).  In particular, clients MUST support both
```
robstradling: This "In particular" sentence draws attention to 2 of the 3 mandatory mechanisms.  Why is the 3rd mechanism less worthy of emphasis?
I think this sentence should either be dropped, or it should enumerate all 3 mandatory mechanisms.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1626,12 +1633,17 @@ client uses `get-all-by-hash`, then it will already have the new STH.
 
 ### Evaluating compliance
 
-To be considered compliant, a certificate MUST be accompanied by at least one
-valid SCT. A certificate not accompanied by any valid SCTs MUST NOT be
-considered compliant by TLS clients.
+To be considered compliant with this specification, a certificate MUST be
```
robstradling: IIRC, "with this specification" was present in a previous iteration of the document, but we took the decision to remove it because it's implied.

robstradling: The idea of "at least one valid SCT" was to specify the lowest common denominator level of compliance upon which all local client policies would build.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1626,12 +1633,17 @@ client uses `get-all-by-hash`, then it will already have the new STH.
...
+accompanied by evidence that it has been included in one or more logs. There
+are two types of proof that a client might require:
```
robstradling: Use "evidence", or use "proof".  I think it's confusing to use two terms that mean the same thing.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1626,12 +1633,17 @@ client uses `get-all-by-hash`, then it will already have the new STH.
...
 
-A TLS client MUST NOT evaluate compliance if it did not send both the
-`transparency_info` and `status_request` TLS extensions in the ClientHello.
+1. A log inclusion proof to a Signed Tree Head
+2. A Signed Certificate Timestamp
+
+It is up to a client's local policy which forms of evidence are sufficient for
+compliance.  For example, a client might require certificates to come with some
+minimum number of SCTs from among a collection of trusted logs, or require a
+single inclusion proof to an audited tree head.
```
robstradling: "SCTs...or...inclusion proof"?  AIUI, to verify inclusion a client needs an inclusion proof, an STH, the certificate *and* the corresponding SCT.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1626,12 +1633,17 @@ client uses `get-all-by-hash`, then it will already have the new STH.
...
+To be considered compliant with this specification, a certificate MUST be
+accompanied by evidence that it has been included in one or more logs. There
+are two types of proof that a client might require:
 
-A TLS client MUST NOT evaluate compliance if it did not send both the
-`transparency_info` and `status_request` TLS extensions in the ClientHello.
+1. A log inclusion proof to a Signed Tree Head
+2. A Signed Certificate Timestamp
+
+It is up to a client's local policy which forms of evidence are sufficient for
```
robstradling: Since you're also proposing that we remove the "Handling of Non-compliance" section, is it also worth explicitly stating here that it's up to the client to decide how to handle non-compliance?

Maybe reword something like this...
"It is entirely up to a client's local policy to specify:
  - which forms and quantity of evidence are considered sufficient to achieve compliance.
  - how to handle non-compliance."

robstradling:
Closing in favour of #266.

PR #245, opened 2017-04-07
Ticket 188: Empty consistency proof for equal tree sizes.

eranmes:
When get-consistency-proof is called with identical `first`, `second`, then the returned consistency proof should be empty.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1216,6 +1216,9 @@ Outputs:
   : A base64 encoded `TransItem` of type `consistency_proof_v2`, whose
     `tree_size_1` MUST match the `first` input. If the `sth` output is omitted,
     then `tree_size_2` MUST match the `second` input.
+    If `first` and `second` are equal and correspond to a known STH, the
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1104,6 +1104,8 @@ Outputs:
   : A base64 encoded `TransItem` of type `consistency_proof_v2`, whose
     `tree_size_1` MUST match the `first` input. If the `sth` output is omitted,
     then `tree_size_2` MUST match the `second` input.
+    If `first` and `second` are equaly, the returned consistency proof MUST
```
robstradling: I suggest: "s/are equaly,/are equal and correspond to a known STH,/"

If first and second are equal but don't correspond to a known STH, then I think the "first unknown" error would be a more appropriate response than an empty "consistency_path".

eranmes: Note that I think it should be OK to return an empty consistency path if the log implementation chooses to allow getting inclusion proofs between arbitrary tree sizes.

PR #246, opened 2017-05-02
Issue #166, log parameters: Richard's PR with reference fixes

eranmes:
This PR is based on https://github.com/google/certificate-transparency-rfcs/pull/226, with a merge commit that adds the anchor for the Log Parametrs section, so existing references do not break.


eranmes:
@bifurcation can you please confirm to the googlebot that you're OK with your commit being contributed via this PR?

robstradling:
LGTM insofar as it goes.  However...
  - We still talk about a log's "metadata" in various other places in the document.  We should fix that.
  - The Clients section still says "This metadata is described below", which refers to the "Metadata" section that (once this PR is merged) will no longer be "below".

eranmes:
@robstradling addressed both points.

eranmes:
@robstradling PTAL, I've changed 'metadata' to 'parameters' throughout and removed the relative reference (that's what we have hyperlinks for!).

PR #238, opened 2017-04-03
Issue #181 - Consolidate 'add-chain' and 'add-pre-chain' endpoints

bifurcation:

PR #181, opened 2016-07-25
Bugfix: Only prefix hashed values with length

eranmes:
Saying that || denotes length-encoded concatenation would apply length-encoding
to the result of LABELHASH and the result of BASE32 (as concatenation is used
3 times in the REDACT function), but length encoding is only necessary for
the values that are being hashed.

PR #248, opened 2017-05-04
Issue #181 - Consolidate 'add-chain' and 'add-pre-chain'

eranmes:
This is a different take on bifurication's pull request (PR #238):
* The certificate submitted has to be explicitly specified in
a separate parameter, rather than implied from the presence of the
precertificate parameter.
* An error code has been added to indicate if a submission is invalid
because both parameters have been specified.

I chose this approach as it explicitly indicates what's the certificate
the submitter wishes to be included in the log, avoiding the potential
issue of logging partial chains because the precertificate was omitted.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -610,7 +610,7 @@ submit them.
 ## Precertificates    {#precertificates}
 
 CAs may preannounce a certificate prior to issuance by submitting a
-precertificate ({{add-pre-chain}}) that the log can use to create an entry that
+precertificate ({{add-chain}}) that the log can use to create an entry that
 will be valid against the issued certificate. The CA MAY incorporate the
 returned SCT in the issued certificate. One example of where the returned SCT is
 not incorporated in the issued certificate is when a CA sends the precertificate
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1156,27 +1164,8 @@ If the certificate is logged, an SCT MUST be issued. Logging the certificate is
...
-
-Outputs:
-
-: sct:
-  : A base64 encoded `TransItem` of type `precert_sct_v2`, signed by this log,
-    that corresponds to the submitted precertificate.
-
-Errors are the same as in {{add-chain}}.
+A log MUST reject a submission where both the `precertificate` and `certificate`
+parameters were specified and return the `ambiguous submission` error.
```
robstradling: What happens if neither `precertificate` nor `certificate` is specified?  Perhaps add a "bad submission" error code to handle that case?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1117,34 +1117,42 @@ modification at a later date. Note that as per [RFC7231], in the case of a 503
...
 
-: chain:
-  : An array of base64 encoded certificates. The first element is the
-    certificate for which the submitter desires an SCT; the second certifies the
+: precertificate:
+  : The base64 encoded precertificate.
+
+  certificate:
+  : The base64 encoded certificate.
+
```
robstradling: I suggest adding a note here to say that either `precertificate` or `certificate` needs to be supplied.  Not both, and not neither.

robstradling:
I prefer the approach in this PR (rather than the approach in PR #238).

draft-ietf-trans-rfc6962-bis.md

eranmes:
@robstradling PTAL.

robstradling:
@eranmes PTAL at PR #262, which incorporates this PR and implements some further changes that were proposed on the list.  (If you're happy with #262, please close this PR).

eranmes:
Rob's change looks better, closing this one.

PR #249, opened 2017-05-04
Follow-up PR for dealing with BCP 190

eranmes:
Follow up to PR #239 , which lacked a proper reference and with another change requested by @robstradling .


eranmes:
@robstradling if you approve of this change I'll land the other PR first, rebase this on master and push it immediately afterwards.

PR #252, opened 2017-05-09
Remove remnants of TLS Feature support.

robstradling:
PR #240 missed some bits.


eranmes:
Thanks for following up, LGTM.

PR #254, opened 2017-05-09
Issue 167: Defining incorporation

eranmes:
Define what it means to incorporate a submission into the tree.

Additionally, clarify what a log MUST do to adhere with this
requirement:
Rather than dictate that this is done periodically, state it MUST
be done within the MMD for each submission, pointing out that batching
of the operation as an implementation detail.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -656,10 +656,15 @@ previously logged as a precertificate, then the precertificate's SCT of type
 `precert_sct_v2` would not be appropriate; instead, a fresh SCT of type
 `x509_sct_v2` should be generated.
 
-An SCT is the log's promise to incorporate the submitted entry in its Merkle
-Tree no later than a fixed amount of time, known as the Maximum Merge Delay
-(MMD), after the issuance of the SCT. Periodically, the log MUST append all its
-new entries to its Merkle Tree and sign the root of the tree.
+An SCT is the log's promise to append to its Merkle Tree an entry for the
+accepted submission. Upon producing an SCT, the log MUST fulfil this promise by
+performing the following actions within a fixed amount of time known as the
+Maximum Merge Delay (MMD), which is one of the log's parameters (see
+{{log_parameters}}):
+* Allocate a tree index to the entry representing the accepted submission.
+* Calculate the root of the tree.
+* Sign the root of the tree (see {{sth}}).
+The log may append multiple entries before signing the root of the tree.
 
 Log operators SHOULD NOT impose any conditions on retrieving or sharing data
 from the log.


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -656,10 +656,15 @@ previously logged as a precertificate, then the precertificate's SCT of type
 `precert_sct_v2` would not be appropriate; instead, a fresh SCT of type
 `x509_sct_v2` should be generated.
 
-An SCT is the log's promise to incorporate the submitted entry in its Merkle
-Tree no later than a fixed amount of time, known as the Maximum Merge Delay
-(MMD), after the issuance of the SCT. Periodically, the log MUST append all its
-new entries to its Merkle Tree and sign the root of the tree.
+An SCT is the log's promise to append, to its Merkle Tree, an entry for the
+submission, no later than a fixed amount of time (known as the Maximum Merge
+Delay, MMD), after the issuance of the SCT. An entry for a submission is
```
robstradling: My attempt at rewording this sentence with fewer commas:
"An SCT is the log's promise to append to its Merkle Tree an entry for the accepted submission no later than a fixed amount of time, known as the Maximum Merge Delay (MMD), after the issuance of the SCT."

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -656,10 +656,15 @@ previously logged as a precertificate, then the precertificate's SCT of type
...
+considered as included in the tree after a tree index was allocated to it
```
robstradling: s/included in/appended to/
s/was/has been/

draft-ietf-trans-rfc6962-bis.md

draft-ietf-trans-rfc6962-bis.md

eranmes:
@robstradling PTAL.

robstradling:
@eranmes How about rewriting the paragraph as follows...
```
An SCT is the log's promise to append to its Merkle Tree an entry for the
accepted submission. Upon producing an SCT, the log MUST fulfil this promise by
performing the following actions within a fixed amount of time known as the
Maximum Merge Delay (MMD), which is specified as one of the log's parameters
(see {{log_parameters}}):
* Allocate a tree index to the accepted submission.
* Calculate the root of the tree.
* Sign the root of the tree (see {{sth}}).
* Publish the root of the tree (see {{get-sth}}).
The log may append multiple entries before signing the root of the tree.
```

eranmes:
While overall looking good, the requirement of publishing a root that includes the entry within the MMD has the subtle effect of halving the MMD: Until now, logs were required to produce an STH that includes a given submission within the MMD, and publish an STH that's no older than the MMD.

That means that an observer watching a log that publishes STHs that are slightly less than MMD behind the current time, will only see an STH covering a particular submission after 2*MMD.

Your change requires creation of an STH and publishing it within an MMD of submission - I'm still on the fence regarding the desirability of that  one (because log front-end skew makes it hard to coordinate STH serving).

robstradling:
@eranmes Good point.  Please drop the "Publish the root of the tree" line I suggested.

Elsewhere in the document we already say "When a client requests a log's latest STH (see {{get-sth}}), the log MUST return an STH that is no older than the log's MMD", so what I was attempting (and failing) to express is already covered.

eranmes:
@robstradling done, PTAL.

PR #256, opened 2017-05-11
Removing X509ChainEntry/PrecertChainEntryV2

eranmes:
These structures are not needed if the value returned by get-entries is the original submission JSON, which will be easier for clients to parse anyway.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -757,41 +757,14 @@ the submitter, and the trust anchor used to verify the chain (even if it was
 omitted from the submission). The log MUST present this chain for auditing upon
 request (see {{get-entries}}). This chain is required to prevent a CA from
 avoiding blame by logging a partial or empty chain.
-
-Each certificate entry in a log MUST include a `X509ChainEntry` structure, and
-each precertificate entry MUST include a `PrecertChainEntryV2` structure:
-
-~~~~~~~~~~~
-    opaque ASN.1Cert<1..2^24-1>;
-
-    struct {
-        ASN.1Cert leaf_certificate;
-        ASN.1Cert certificate_chain<0..2^24-1>;
-    } X509ChainEntry;
-
-    opaque CMSPrecert<1..2^24-1>;
-
-    struct {
-        CMSPrecert pre_certificate;
-        ASN.1Cert precertificate_chain<1..2^24-1>;
-    } PrecertChainEntryV2;
-~~~~~~~~~~~
-
-`leaf_certificate` is a submitted certificate that has been accepted by the log.
-
-`certificate_chain` is a vector of 0 or more additional certificates required to
-verify `leaf_certificate`. The first certificate MUST certify
-`leaf_certificate`. Each following certificate MUST directly certify the one
-preceding it. The final certificate MUST be a trust anchor accepted by the log.
-If `leaf_certificate` is an accepted trust anchor, then this vector is empty.
-
-`pre_certificate` is a submitted precertificate that has been accepted by the
-log.
-
-`precertificate_chain` is a vector of 1 or more additional certificates required
-to verify `pre_certificate`. The first certificate MUST certify
-`pre_certificate`. Each following certificate MUST directly certify the one
-preceding it. The final certificate MUST be a trust anchor accepted by the log.
+Log entries, which are the input to the Merkle Tree, are TLS-encoded
+`TransItem`s of type `x509_entry_v2` or `precert_entry_v2`.
+To be able to construct a `TransItem` representing the entry in the future, in
+addition to storing the original submission, the log also has to store the
+timestamp of the issued SCT, as well as any SCT extensions.
+Implementations may store all of this information in any format, but MUST
+be able to construct from it the corresponding log entries, in the order they
+were created and added to the tree.
 
 ## Log ID    {#log_id}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -751,41 +751,13 @@ the submitter, and the trust anchor used to verify the chain (even if it was
...
-`pre_certificate` is a submitted precertificate that has been accepted by the
-log.
-
-`precertificate_chain` is a vector of 1 or more additional certificates required
-to verify `pre_certificate`. The first certificate MUST certify
-`pre_certificate`. Each following certificate MUST directly certify the one
-preceding it. The final certificate MUST be a trust anchor accepted by the log.
+Log entries, which are the input to the Merkle Tree, are TLS-encoded
+`TransItem`s of type `x509_entry_v2` or `precert_entry_v2`.
+To construct those, the log MUST store the timestamp of the issued SCT, as well
```
ln5: It seems wrong to have normative language about storage. What about something like the following?

    To be able to construct a `TransItem` in the future, in addition to storing the original submission, the log also has to store the timestamp of the issued SCT, as well as any SCT extensions.

eranmes:
@robstradling rebased on top of your submit-entry PR , PTAL.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -757,41 +757,14 @@ the submitter, and the trust anchor used to verify the chain (even if it was
...
-`pre_certificate`. Each following certificate MUST directly certify the one
-preceding it. The final certificate MUST be a trust anchor accepted by the log.
+Log entries, which are the input to the Merkle Tree, are TLS-encoded
+`TransItem`s of type `x509_entry_v2` or `precert_entry_v2`.
+To be able to construct a `TransItem` representing the entry in the future, in
+addition to storing the original submission, the log also has to store the
+timestamp of the issued SCT, as well as any SCT extensions.
+Implementations may store all of this information in any format, but MUST
+be able to construct from it the corresponding log entries, in the order they
+were created and added to the tree.
```
robstradling: I think "in the order they were created and added to the tree" is out of place in this section.  We haven't discussed adding entries to the tree yet.

robstradling: I suggest deleting this sentence, and then using the following text (note: this incorporates my other wordsmithing attempts during this review)...

"Each log entry is a `TransItem` structure of type `x509_entry_v2` or `precert_entry_v2`. However, a log may store its entries in any format. If a log does not store each `TransItem` in full, it must at least store the `timestamp` and `sct_extensions` of the corresponding `TimestampedCertificateEntryDataV2` structure. Armed with these fields and the entire chain used to verify each submission, a log will be able to reconstruct each `TransItem`."

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1411,14 +1384,16 @@ Outputs:
 : entries:
   : An array of objects, each consisting of
 
-    leaf_input:
+    log_entry:
     : The base64 encoded `TransItem` structure of type `x509_entry_v2` or
       `precert_entry_v2` (see {{tree_leaves}}).
```
robstradling: Why reference {{tree_leaves}} here?  Wouldn't it be more appropriate to reference {{log_entries}} ?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1411,14 +1384,16 @@ Outputs:
...
-    log_entry:
-    : The base64 encoded log entry (see {{log_entries}}). In the case of an
-      `x509_entry_v2` entry, this is the whole `X509ChainEntry`; and in the case
-      of a `precert_entry_v2`, this is the whole `PrecertChainEntryV2`.
+    submission:
+    : JSON object representing the original certificate, or precertificate,
+      submission. In the case of an `x509_entry_v2` entry, the object would
+      contain `certificate` and `chain` fields; ; and in the case
+      of a `precert_entry_v2`, it would contain `precertificate` and `chain`
+      fields.
```
robstradling: We've changed the "submit-entry" API to accept inputs named "submission" and "type".  We should make the same change here, right?

Actually, I suggest that you:
1. rename the "submission" output to "submitted_entry" (to mirror the API name "submit-entry").
2. simplify the description to: "JSON object that contains the inputs that were submitted to `submit-entry`."

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1441,6 +1416,10 @@ as returned by `get-sth` in {{get-sth}}.
 
 The `start` parameter MUST be less than or equal to the `end` parameter.
 
+The `chain` field in the `submission` output parameter MUST include the trust
+anchor that the submission chained to, even if it was omitted in the original
```
robstradling: I suggest: s/that the submission chained to/that the log used to verify the submission/

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -757,41 +757,14 @@ the submitter, and the trust anchor used to verify the chain (even if it was
...
-If `leaf_certificate` is an accepted trust anchor, then this vector is empty.
-
-`pre_certificate` is a submitted precertificate that has been accepted by the
-log.
-
-`precertificate_chain` is a vector of 1 or more additional certificates required
-to verify `pre_certificate`. The first certificate MUST certify
-`pre_certificate`. Each following certificate MUST directly certify the one
-preceding it. The final certificate MUST be a trust anchor accepted by the log.
+Log entries, which are the input to the Merkle Tree, are TLS-encoded
```

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -757,41 +757,14 @@ the submitter, and the trust anchor used to verify the chain (even if it was
...
+`TransItem`s of type `x509_entry_v2` or `precert_entry_v2`.
+To be able to construct a `TransItem` representing the entry in the future, in
+addition to storing the original submission, the log also has to store the
+timestamp of the issued SCT, as well as any SCT extensions.
```
eranmes: I've re-worded the paragraph based on your suggestion (but not exactly it).

draft-ietf-trans-rfc6962-bis.md

draft-ietf-trans-rfc6962-bis.md
eranmes: Done, after adding the caveat that the JSON objects contains the entire chain, even if the trust anchor was missing from the submission.

eranmes: This is mentioned below but I've mentioned it here again to avoid having two conflicting statements.

draft-ietf-trans-rfc6962-bis.md

eranmes:
@robstradling PTAL.

PR #257, opened 2017-05-12
Issue 175: Clarifying guarantees around MMD, STH age.

eranmes:
Clarify that it make take up to twice the MMD for a submission
to be available to log monitors, since a submission must be
incorporated within the MMD, but an old STH may be served so
the incorporated entry may not be available until the STH that
includes it is served.


robstradling:
I think it might be better to incorporate this paragraph into the "Misissued Certificates" section that immediately follows.

That section currently says...
`Thus, the maximum period of time during which a misissued certificate can be used without being available for audit is the MMD.`
...which, IINM, is at odds with the paragraph that this PR wants to add.

eranmes:
@robstradling PTAL, amended the following section (instead of adding text) as you suggested.

PR #164, opened 2016-05-23
Grammar nit.

wkumari:
[O] to one or more logs, however certificate
[P] to one or more logs; however, certificate
[R] Grammar/run on sentence

No formatting change. Original line: 125 words, 742 characters.
Long lines make github diff sad.

PR #258, opened 2017-05-12
Issue #164: Clarify roles of TLS clients.

eranmes:
Remove repeated text about TLS clients having to validate inclusion
proofs.
Make it clear they SHOULD do it, though.


eranmes:
@robstradling implemented your suggestion, PTAL.

PR #261, opened 2017-05-22
Remove deterministic signatures requirement

eranmes:
Per discussion on trans, making deterministic signing optional,
mentioning that deterministic signing can be achieved by storing
the signature blobs rather than re-signing the same SCTs for
the same submissions.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -2072,15 +2072,17 @@ the log, proving violation of the append-only property.
 
 ## Deterministic Signatures    {#deterministic_signatures}
 
-Logs are encouraged to use deterministic signatures to reduce the risk of
-tracking clients: clients that gossip STHs or report back SCTs can be tracked
-or traced if a log was to produce multiple STHs or SCTs with the same timestamp
-and data but different signatures.
+Clients that gossip STHs or report back SCTs can be tracked or traced if a log
+produces multiple STHs or SCTs with the same timestamp and data but different
+signatures.
+Logs SHOULD mitigate this risk by either:
 
-Note that deterministic signatures can be provided without the use of a
-signature scheme that's designed for deterministic signing: The signature bytes
-for each SCT or STH could be stored and served when the same certificate is
-submitted or the same STH requested, rather than re-signing the same data.
+- Using deterministic signature schemes, or
+
+- Producing no more than one SCT for each distinct submission and no more than one
+STH for each distinct tree_size. Each of these SCTs and STHs can be stored by
+the log and served to other clients that submit the same certificate or request
+the same STH.
 
 ## Multiple SCTs    {#offering_multiple_scts}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -2067,15 +2065,15 @@ the log, proving violation of the append-only property.
 
 ## Deterministic Signatures    {#deterministic_signatures}
 
-Logs are required to use deterministic signatures for the following reasons:
+Logs are encouraged to use deterministic signatures to reduce the risk of
+tracking clients: clients that gossip STHs or report back SCTs can be tracked
+or traced if a log was to produce multiple STHs or SCTs with the same timestamp
+and data but different signatures.
```
robstradling: My wordsmithing attempt on this paragraph:
`Clients that gossip STHs or report back SCTs can be tracked or traced if a log produces multiple STHs or SCTs with the same timestamp and data but different signatures.  To mitigate this risk, logs are encouraged to use deterministic signature schemes.`

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -2067,15 +2065,15 @@ the log, proving violation of the append-only property.
...
-  that each signature can potentially expose the secret material of the signing
-  key.
-
-* Clients that gossip STHs or report back SCTs can be tracked or traced if a log
-  was to produce multiple STHs or SCTs with the same timestamp and data but
-  different signatures.
+Note that deterministic signatures can be provided without the use of a
+signature scheme that's designed for deterministic signing: The signature bytes
+for each SCT or STH could be stored and served when the same certificate is
+submitted or the same STH requested, rather than re-signing the same data.
```
robstradling: I think it's misleading/confusing to use the phrase "deterministic signatures can be provided" here.  Here's an attempt at wordsmithing both this and the preceding paragraph:

`Clients that gossip STHs or report back SCTs can be tracked or traced if a log produces multiple STHs or SCTs with the same timestamp and data but different signatures. Logs MUST mitigate this risk by either:
* Using deterministic signature schemes, or
* Producing no more than one SCT for each distinct submission and no more than one STH for each distinct tree_size. Each of these SCTs and STHs can be stored by the log and served to other clients that submit the same certificate or request the same STH.`

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1892,7 +1892,7 @@ IANA is asked to establish a registry of signature algorithm values, named
 |-------------+--------------------------------------------------------+------------------------------------------|
 | Value       | Signature Algorithm                                    | Reference / Assignment Policy            |
 |-------------+--------------------------------------------------------+------------------------------------------|
-| 0x00        | Deterministic ECDSA (NIST P-256) with HMAC-SHA256      | [RFC6979]                                |
+| 0x00        | ECDSA (NIST P-256) with HMAC-SHA256                    | [RFC6979]                                |
```
HMAC-SHA256 seems enough, by looking at the Building Blocks section of RFC6979 (3.2).

draft-ietf-trans-rfc6962-bis.md

draft-ietf-trans-rfc6962-bis.md
eranmes: Done - with one modification, changing MUST to a SHOULD because I really don't want to require it yet, we don't really know how gossip will take place and if it'll indeed introduce a new tracking/fingerprinting vector.

eranmes:
@robstradling PTAL.

draft-ietf-trans-rfc6962-bis.md
robstradling: OK

PR #262, opened 2017-05-25
Issue #181 - Consolidate 'add-chain' and 'add-pre-chain'

robstradling:
This PR incorporates and is intended to replace PRs #238 and #248.  As per discussion on the list...
* the consolidated endpoint is now called "submit-entry" (rather than "add-entry").
* there's now only one parameter ("submission") that carries the submitted certificate or precertificate.

I've also updated the parameter descriptions and error messages accordingly.


eranmes:
LGTM - how do you envision that interacting with the changes in PR #256 ?

PR #263, opened 2017-06-15
Issue #189: Permit logs to use EdDSA.  Also, drop RSA.

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -54,6 +54,7 @@ normative:
   RFC7633:
   RFC7924:
   RFC8017:
```
eranmes: Drop the reference to RFC8017, since it's not used anymore AFAICT.

PR #79, opened 2015-08-19
Initial check-in for ID for DNS

aeijdenberg:
It's possibly not quite in a great condition to check-in, but I may be out for a few days so wanted to get something out.

PR #264, opened 2017-06-16
Issue #79 revisited - Precertificate signature must be over something other than just the TBSCertificate

robstradling:
Explain how a precertificate signature cannot be a certificate signature.
Draw attention to the RFC5652 requirement to include the 'signedAttrs' field.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -649,11 +649,11 @@ following requirements:
 
 As described in Section 5.3 of [RFC5652], the `SignerInfo.signedAttrs` field
 MUST be present (because the `eContentType` is not `id-data`) and MUST contain
-at least a content-type attribute and a message-digest attribute. This field is
-included in the message digest calculation process (see Section 5.4 of
-[RFC5652]), which ensures that the `signature` in the CMS object will not be a
-valid X.509v3 signature and so cannot be used to construct a certificate from
-the precertificate.
+a content-type attribute and a message-digest attribute. This field is included
+in the message digest calculation process (see Section 5.4 of [RFC5652]), which
+ensures that the `signature` in the CMS object will not be a valid X.509v3
+signature and so cannot be used to construct a certificate from the
+precertificate.
 
 # Log Format and Operation
 
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -643,6 +643,8 @@ following requirements:
   is considered binding (i.e., misissuance of the precertificate is considered
   equivalent to misissuance of the corresponding certificate).
 
+* `SignerInfo.SignerIdentifier` MUST use the `subjectKeyIdentifier` option.
+
 * `SignedData.certificates` SHOULD be omitted.
 
 As described in Section 5.3 of [RFC5652], the `SignerInfo.signedAttrs` field
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -647,13 +647,20 @@ following requirements:
 
 * `SignedData.certificates` SHOULD be omitted.
 
-As described in Section 5.3 of [RFC5652], the `SignerInfo.signedAttrs` field
-MUST be present (because the `eContentType` is not `id-data`) and MUST contain
-a content-type attribute and a message-digest attribute. This field is included
-in the message digest calculation process (see Section 5.4 of [RFC5652]), which
-ensures that the `signature` in the CMS object will not be a valid X.509v3
-signature and so cannot be used to construct a certificate from the
-precertificate.
+Since the `eContentType` is not `id-data`, the `SignerInfo.signedAttrs` field
+MUST be present (see Section 5.3 of [RFC5652]) and MUST contain:
+
+* A content-type attribute whose value is the same as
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -637,16 +637,22 @@ following requirements:
...
   MUST be omitted.
 
-* `SignedData.signerInfos` MUST contain a signature from the same (root or
-  intermediate) CA that will ultimately issue the certificate. This signature
-  indicates the CA's intent to issue the certificate. This intent is considered
-  binding (i.e., misissuance of the precertificate is considered equivalent to
-  misissuance of the certificate). (Note that, because of the structure of CMS,
-  the signature on the CMS object will not be a valid X.509v3 signature and so
-  cannot be used to construct a certificate from the precertificate).
+* `SignedData.signerInfos` MUST contain 1 `SignerInfo` with a `signature` from
```
eranmes: Nit: I notice the SignerInfo can identify the signer either by issuer and serial number (version = 1) or the subjectKeyIdentifier (version = 3).
Do we want to limit this further by requiring that the signer is identified by subjectKeyIdentifier?

eranmes:
LGTM - having read sections 5.3 and 5.4 rfc5652, I understand that the signedAttrs are included in the message digest and so the signature is not only over the TBSCertificate bytes.

(It's still unclear to me how exactly the signedAttrs DER encoding get concatenated to the encapContentInfo eContent)

draft-ietf-trans-rfc6962-bis.md
robstradling: I like the idea of limiting this further, because that would make log implementation a bit simpler.

However, that would mean overriding this requirement in https://tools.ietf.org/html/rfc5652#section-5.3...
`      Implementations MUST support the reception of the
      issuerAndSerialNumber and subjectKeyIdentifier forms of
      SignerIdentifier.`
Can we do that?

eranmes: I think we do - the part on the log side doesn't have to be a fully-compliant RFC5652 receiver, I think there are plenty of other attributes it will ignore.

We can specify that the sender (producer of the CMS precert) must use the subjectKeyIdentifier approach and no other signedAttrs on unsignedAttrs, so that means the implementation on the log side doesn't have to care about any of those other than what's specified here.

robstradling:
I agree that it's unclear how the concatenation is done.  https://tools.ietf.org/html/rfc5652#section-5.4 says...
`In either case, the initial input to the message digest calculation process is the "value" of the encapsulated content being signed'
...but then it seems to disregard that "initial input" when it says (referring to the signedAttrs field)...
`When the field is present, however, the result is the message digest of the complete DER encoding of the SignedAttrs value contained in the signedAttrs field.`

My guess is that the intent is to append the "complete DER encoding of the SignedAttrs value" after the "initial input", and then calculate the message digest of that.

Since it's unclear, I think we may need to file an erratum for RFC5652.  But first, I'll reach out to Russ Housley (RFC5652 author) to discuss this.

robstradling:
I've asked the LAMPS WG to comment:
https://mailarchive.ietf.org/arch/msg/spasm/ACveftefCNw2KYeU-bdgZA64Y5E

robstradling:
Russ Housley posted some pseudo-code [1] to clarify how it works.  Two message digests are computed:
- The digest of the TBSCertificate (which is placed into a message-digest attribute, which is then placed into the SignedAttrs).
- The digest of the SignedAttrs.

[1] https://mailarchive.ietf.org/arch/msg/spasm/mg2-u8sfnkuQa8ZoAMM3X9rJhc8

eranmes:
Great, is there anything else needed before this PR can be merged?

draft-ietf-trans-rfc6962-bis.md
robstradling: Done.

robstradling:
@eranmes PTAL.  Do the last two commits address your nits adequately?

robstradling:
@eranmes I've pushed another 6 commits that more fully specify the precertificate requirements and further restrict the profile.  PTAL.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -626,26 +626,48 @@ not incorporated in the issued certificate is when a CA sends the precertificate
...
+
+* `SignedData.signerInfos` MUST contain one `SignerInfo`:
+  * `version` MUST be v3(3).
+  * `sid` MUST use the `subjectKeyIdentifier` option.
+  * `digestAlgorithm` MUST be one of the hash algorithm OIDs listed in
+    {{hash_algorithms}}.
+  * `signedAttrs` MUST be present and MUST contain two attributes:
+    * A content-type attribute whose value is the same as
+      `SignedData.encapContentInfo.eContentType`.
+    * A message-digest attribute whose value is the message digest of
```
eranmes: RFC5652 is rather unclear about whether the message-digest attribute is implicitly added when calculating the message digest (i.e. it doesn't have to be included in the signedAttrs but added when the message digest is calculated) or have to be explicitly present, but it seems like  it has to be explicitly present.

robstradling: I think RFC5652 section 5.3 is clear that the message-digest attribute has to be explicitly present in signedAttrs:
`signedAttrs...If the field is present, it MUST contain, at a minimum...A message-digest attribute.`

PR #267, opened 2017-06-29
Add optional 'sth' and 'inclusion' outputs to submit-entry.

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1199,6 +1199,18 @@ Outputs:
   : A base64 encoded `TransItem` of type `x509_sct_v2` or `precert_sct_v2`,
     signed by this log, that corresponds to the `submission`.
 
+  If the submitted entry has already been appended to this log's tree, or if
```
eranmes: nit: 'has already been appended' -> 'already exists' ?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1199,6 +1199,18 @@ Outputs:
...
+
+  sth:
+  : A base64 encoded `TransItem` of type `signed_tree_head_v2`, signed by this
+    log.
+
+  inclusion:
+  : A base64 encoded `TransItem` of type `inclusion_proof_v2` whose
+    `inclusion_path` array of Merkle Tree nodes proves the inclusion of the
+    `submission` in the returned `sth`.
+
```
eranmes: nit: Add a sentence saying:
If the obtained SCT is meant for use by TLS clients, the `sth` and `inclusion` outputs SHOULD be made available together with the SCT (for example, by embedding all of them in the final certificate).

eranmes:
overall looks good, two minor comments.

draft-ietf-trans-rfc6962-bis.md
robstradling: I'll reword that sentence as follows:
`If the submitted entry is immediately appended to, or already exists in, this log's tree, then the log SHOULD also output:`

draft-ietf-trans-rfc6962-bis.md
robstradling: Good idea.  I've added this sentence:
"If the returned `sct` is intended to be provided to clients, then `sth` and
`inclusion` (if returned) SHOULD also be provided to clients (e.g., if `type`
was 1 then all three `TransItem`s could be embedded in the certificate)."

robstradling:
@eranmes Comments addressed.  PTAL.

PR #272, opened 2017-07-18
Small editorial fixes

eranmes:
* Use the right output parameter name.
* Mention replacement of add-chain / add-pre-chain with submit-entry.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -239,6 +239,8 @@ community. The major changes are:
 - get-all-by-hash: new client API for obtaining an inclusion proof and the
   corresponding consistency proof at the same time.
 
+- submit-entry: new client API, replacing add-chain and add-pre-chain.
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1447,9 +1449,9 @@ as returned by `get-sth` in {{get-sth}}.
 
 The `start` parameter MUST be less than or equal to the `end` parameter.
 
-The `chain` field in the `submission` output parameter MUST include the trust
-anchor that the log used to verify the submission, even if it was omitted in the
-original submission.
+The `chain` field in the `submitted_entry` output parameter MUST include the
```
robstradling: s/in the/in each/

(There's a separate `submitted_entry` output parameter for each element in the `entries` output parameter).

eranmes: Good point - fixed.

PR #273, opened 2017-07-25
Clarify that non-self-signed trust anchors cannot be submitted to submit-entry.

robstradling:
Mention mutation of certificate signatures as a log misbehaviour.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1193,7 +1193,7 @@ Inputs:
   : An array of zero or more base64 encoded CA certificates. The first element
     is the signer of the `submission`; the second certifies the first; etc.
     The last element of `chain` (or, if `chain` is an empty array, the
-    `submission`) either is, or is certified by, an accepted trust anchor.
+    `submission`) is certified by an accepted trust anchor.
 
 Outputs:
 
@@ -1235,6 +1235,11 @@ the log MUST either log the certificate or return the "bad certificate" error.
 If the certificate is logged, an SCT MUST be issued. Logging the certificate is
 useful, because monitors ({{monitor}}) can then detect these encoding errors,
 which may be accepted by some TLS clients.
+
+If `submission` is an accepted trust anchor whose certifier is neither an
+accepted trust anchor nor the first element of `chain`, then the log MUST return
+the "unknown anchor" error. A log cannot generate an SCT for a submission if it
+does not have access to the issuer's public key.
 
 If the returned `sct` is intended to be provided to TLS clients, then `sth` and
 `inclusion` (if returned) SHOULD also be provided to TLS clients (e.g., if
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1236,6 +1236,11 @@ If the certificate is logged, an SCT MUST be issued. Logging the certificate is
 useful, because monitors ({{monitor}}) can then detect these encoding errors,
 which may be accepted by some TLS clients.
 
+If `submission` is an accepted trust anchor whose certifier is neither an
+accepted trust anchor nor the first element of `chain`, then the log MUST return
```
eranmes: Is it clear, though, that a submission of an accepted trust anchor shouldn't be accepted unless there's *another* accepted trust anchor in the chain? Could anyone mis-interpret this section to mean that a non-self-signed accepted trust anchor may be submitted with its chain and it would still be accepted because it contains a trust anchor?

robstradling: I don't think so.  It says "whose certifier".  A non-self-signed accepted trust anchor is not its own certifier, by definition.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -2055,8 +2060,9 @@ action when a misissue is detected.
 A log can misbehave in several ways. Examples include: failing to incorporate a
 certificate with an SCT in the Merkle Tree within the MMD; presenting different,
 conflicting views of the Merkle Tree at different times and/or to different
-parties; and issuing STHs too frequently. Such misbehavior is detectable and
-[I-D.ietf-trans-threat-analysis] provides more details on how this can be done.
+parties; issuing STHs too frequently; and mutating the signature of a logged
+certificate. Such misbehavior is detectable and [I-D.ietf-trans-threat-analysis]
+provides more details on how this can be done.
```
AGWA: This should also mention "failure to furnish a chain containing the certifier of a logged certificate" as misbehavior.

robstradling: OK, I'll add that to the list of example log misbehaviours.

robstradling:
I think I see what you're wanting to clarify...
`If the first certificate in the "chain" field in the "submission" output parameter is not self-signed, then "chain" MUST contain a second certificate, and the second certificate MUST certify the first certificate.`

...but I think it should say...
`If the "submission" does not certify itself, then the first element of "chain" MUST be present and MUST certify the "submission".`

Does that make sense?

AGWA:
@robstradling Yup, you're right.  I forgot that the logged certificate was separate from the chain.

robstradling:
@eranmes @AGWA PTAL.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1191,9 +1191,9 @@ Inputs:
 
   chain:
   : An array of zero or more base64 encoded CA certificates. The first element
-    is the signer of the `submission`; the second certifies the first; etc.
+    is the certifier of the `submission`; the second certifies the first; etc.
     The last element of `chain` (or, if `chain` is an empty array, the
-    `submission`) either is, or is certified by, an accepted trust anchor.
+    `submission`) is certified by an accepted trust anchor.
```
eranmes: Not sure I recall - why are we no longer allowing inclusion of the trust anchor itself in the submission?

draft-ietf-trans-rfc6962-bis.md
eranmes: I'm still confused by this bit: "... the first element of `chain`":
When is it valid to submit an accepted trust anchor?
It seems to me the answer is:
- When there's another trust anchor that certifies it, or:
- When there's a chain that leads to another trust anchor.
Is that right? Are self-signed trust anchors considered as submissions "whose certifier is a trust anchor" - itself?

robstradling: Yes, a self-signed trust anchor is the certifier of itself.  When submitting a self-signed trust anchor in `submission`, the `chain` can be empty.

draft-ietf-trans-rfc6962-bis.md
robstradling: If the "last element of `chain` (or, if `chain` is an empty array, the `submission`)...is...an accepted trust anchor", and if that accepted trust anchor is not self-signed, then...well...that's what we're trying to disallow in this PR.

We do still allow inclusion of a trust anchor in the submission.  A self-signed trust anchor is certified by itself.

PR #278, opened 2017-10-26
A-D review: Clarify that the submitter does the path construction in the Accepting Submissions section

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -750,7 +750,8 @@ elements.
 To avoid being overloaded by invalid submissions, the log MUST NOT accept any
 submission until it has verified that the submitted certificate or
 precertificate has a valid signature chain to an accepted trust anchor, using
-only the chain of intermediate CA certificates provided by the submitter.
+only the chain of intermediate CA certificates constructed and provided by the
```
eranmes: Is this change to make it clear that the log must not attempt to construct a certificate chain itself from all the intermediates provided by the submitter (if, for example, they were in the wrong order) ?
If so maybe just say "using only the chain of intermediate CA certificates as provided by the submitter, in the order provided" or something similar?
I personally find that adding "constructed" here a bit confusing. Maybe just me.

robstradling: From the TRANS list:
EKR: Hmm... So your theory is that the submitter does path construction
Ben: Yep - and that's borne out by practice.
EKR: OK, well, then it would help if the document were clearer on this point.

I've had another go at making it clearer.

PR #283, opened 2017-10-27
A-D review: Various fixes in the TLS Clients section

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1667,43 +1667,46 @@ their responses. These parameters are described in {{log_parameters}}, but note
...
-mechanisms will be in separate documents, but it is expected there will be a
-variety.
-
 ## TLS Client {#tls_clients}
 
-### Receiving SCTs and inclusion proofs
+### Receiving SCTs and inclusion proofs {#receiving_transitems}
+
+TLS clients receive SCTs and inclusion proofs alongside or in certificates.
+CT-using TLS clients MUST implement all of the three mechanisms by which TLS
```
eranmes: Same nit: ct-using/ct-aware.
I don't mind if we use the term "CT-using" (I'm no native speaker to determine which is more palatable), just that we're consistent.

robstradling: I've gone with "CT-using".

I think "CT-aware" would also include software whose CT functionality has been disabled by the user, whereas "CT-using" can only mean software that is actively using CT.

PR #284, opened 2017-10-27
A-D review: Various fixes in the Monitors section

robstradling:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1761,29 +1761,37 @@ interest, or both. For example, a monitor may be configured to report on all
...
-4. Confirm that the tree made from the fetched entries produces the same hash as
+5. Confirm that the tree made from the fetched entries produces the same hash as
    that in the STH.
 
-5. Fetch the current STH ({{get-sth}}). Repeat until the STH changes.
+To inspect new entries, the monitor SHOULD follow these steps repeatedly for
+each log:
 
-6. Verify the STH signature.
+6. Fetch the current STH ({{get-sth}}). Repeat until the STH changes.
```
eranmes: nit: Re-start numbering from 1? it's not really a part of the same process.

robstradling: I agree it's not part of the same process.  However, if I were to restart the numbering then "11. Repeat from step 6" would become "6. Repeat from step 1" and there would be two step 1s on the page.  Which step 1 would it be referring to?
I'd rather not restart the numbering simply so that this ambiguity is avoided.

eranmes: Fine by me.

PR #289, opened 2017-11-15
A-D review: MUSTs -> SHOULDs.

robstradling:
EKR wrote "I'm not arguing that you shouldn't require submission of complete chains, I'm saying I don't understand why MUST verify is a conformance requirement."

I just chatted to @AlCutter, who convinced me that we should downgrade the MUSTs in this paragraph to SHOULDs.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -753,11 +753,11 @@ elements.
 
 ## Accepting Submissions
 
-To avoid being overloaded by invalid submissions, the log MUST NOT accept any
+To avoid being overloaded by invalid submissions, the log SHOULD NOT accept any
 submission until it has verified that the certificate or precertificate was
 submitted with a valid signature chain to an accepted trust anchor. The log
-MUST NOT use other sources of intermediate CA certificates to attempt
-certification path construction; instead, it MUST only use the intermediate CA
+SHOULD NOT use other sources of intermediate CA certificates to attempt
+certification path construction; instead, it SHOULD only use the intermediate CA
 certificates provided in the submission, in the order provided.
 
 Logs SHOULD accept certificates and precertificates that are fully valid
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -756,8 +756,8 @@ elements.
 To avoid being overloaded by invalid submissions, the log SHOULD NOT accept any
 submission until it has verified that the certificate or precertificate was
 submitted with a valid signature chain to an accepted trust anchor. The log
-SHOULD NOT use other sources of intermediate CA certificates to attempt
-certification path construction; instead, it SHOULD only use the intermediate CA
+MUST NOT use other sources of intermediate CA certificates to attempt
+certification path construction; instead, it MUST only use the intermediate CA
 certificates provided in the submission, in the order provided.
 
 Logs SHOULD accept certificates and precertificates that are fully valid


eranmes:
@robstradling mind sharing the rationale?

eranmes:
In particular this has been a pretty essential requirement to protect the ecosystem from spam in logs. Monitors not having to worry about the validity of each chain in the log makes monitors simpler and generally limits one source of complexity to logs themselves (i.e. only logs have to check submission validity, so monitors can raise the alarm with high confidence if a suspicious cert is encountered, without validating the chain).

RFC2119 says this about SHOULD:

This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

I am struggling to imagine valid reasons to admit into the log an invalid chain in some circumstances.

AlCutter:
I'll play devil's avocado for mo' - let's say, entirely "hypothetically" that there's an intermediate kicking about whose signature is encoded in a technically invalid fashion, or, if you prefer, a root with CA:false, but it's worked-around (or goes blindly unnoticed) by the UA stacks.  Should EE certs which traverse those gems be loggable? I'd argue yes, because I cling tightly onto my "if a UA will trust it then it should be logged" paddle, but having the "MUST verify" wording makes that impossible, for a strict value of verify at least.

Or, someone comes along and claims your venerable & trusted log is now non-compliant with 6962bis and should be distrusted forthwith because:
 - your cert parser/verifier was lax/forgiving in the past and you accepted a cert in something like the above situation  
 - there's a bug in your log's cert validation logic and you accept something which you definitely shouldn't have.
 - you accept a chain which passes through a temporally valid but revoked intermediate you weren't aware of at the time of acceptance (nobody checks this anyway, and even if they did it's a race)
 - etc., etc.
I'd expect they'd be laughed out of court, even though they'd be technically correct with the MUST.

Going ad absurdum in the other direction, who cares if a Log operator wants to allow whatever the hell they like into their log? The economics of the ecosystem are such that they'll find themselves with an expensive to operate Log which no UA will trust because of the cost of monitoring and the likelihood that this log won't make it past its compliance period.

So, in summary:
 - if -bis should be actually deployable that MUST isn't a MUST.
 - there are obvious reasons why the text is RECOMMENDED, and the pre-existing incentives for operators align with that.

eranmes:
This topic has already been discussed on the trans mailing list, at least twice, the latest time being a year ago (https://www.ietf.org/mail-archive/web/trans/current/msg02484.html) and earlier (https://trac.ietf.org/trac/trans/ticket/73, https://trac.ietf.org/trac/trans/ticket/21).

Read section 4.2. further on: The third paragraph explicitly says that "Logs May accept certificates and precertficates that [...] are otherwise not fully valid according to RFC5280 verification rules".

RFC6962-bis already allows accepting of chains that are not fully valid according to RFC5280. The requirement to always check submissions is to ensure logs behave in a consistent, reproducible way according to the protocol. Making submission checking optional means that submission acceptance would be non-deterministic (i.e. a log could be implemented to  just accept all submissions, including invalid ones, in some circumstances, but reject invalid ones on other circumstances).

So the argument doesn't hold - logs can already accept submissions that are not valid according to RFC5280 - they should just so in a deterministic way. The argument that a particular compliance program may provide incentives to operate logs in a certain way does not preclude designing a robust protocol, especially since aforementioned compliance program is outside the IETF.

AGWA:
Strong object to this change.

This requirement is not just for anti-spam, but for security.  Logs need to contain cryptographic proof of which CA misissued a certificate.  If you allow logs to include invalid chains, then a CA could misissue a certificate, collude with a log to log it with an invalid chain, and when someone notices the misissued certificate, the CA can plausibly claim it never issued it.

AlCutter:
> This topic has already been discussed on the trans mailing list, at least twice, the latest time being a year ago (https://www.ietf.org/mail-archive/web/trans/current/msg02484.html) and earlier (https://trac.ietf.org/trac/trans/ticket/73, https://trac.ietf.org/trac/trans/ticket/21).

Due to AD review, it looks like it's being discussed again :/

> Read section 4.2. further on: The third paragraph explicitly says that "Logs May accept certificates and precertficates that [...] are otherwise not fully valid according to RFC5280 verification rules".
> 
> RFC6962-bis already allows accepting of chains that are not fully valid according to RFC5280.

To make sure I understand, your position is that it's OK to log arbitrary DER encoded rubbish so long as it's correctly signed, but it's not OK to log actual issued certificates, that are trusted by UAs, and have a slight encoding error in the representation of the signature?

> The argument that a particular compliance program may provide incentives to operate logs in a certain way does not preclude designing a robust protocol, 

I'm sorry, I don't follow :/
Could you explain how allowing log operators to determine what's allowed into their logs affects the 6962-bis protocol?

Perhaps naively, it seems to me that:
   req: "add this chain, please"
   resp: "Ok, 200 - done." OR  "Nope, 400 - it's borked, go away." OR "500 - I'm borked, try again."
is a robust protocol.

If you're arguing that a "robust protocol" means you should know the add-chain is going to succeed before you send the add-chain, then that's, hopefully obviously, an impossible requirement due to a plethora of reasons including: changing policies, changing root sets, passage of time, bugs, availability, updates to underlying openssl version, networks, downtime, etc. etc. 
Obviously many, if not all, of those reasons also apply to the submitter's assertion that a particular chain is good and therefore should be expected to submit successfully.

> especially since aforementioned compliance program is outside the IETF. 

Then surely the MMD should also defined by -bis, no?
That seems like a pretty important one...

Anyway, getting back to the change, the paragraph Rob's PR affects starts "To avoid being overloaded by invalid submissions, the log MUST NOT ...", so clearly this is not about protocol correctness or what the log should or should not contain. It's also insufficient in its stated purpose; chain checking takes CPU time, which is finite, as is bandwidth to carry the request & response, and without load-shedding, a new log can easily be smashed into a pulp by adding the union of (valid chaining) certs contained in all the existing Logs.

What a log accepts is, along with the implementation and limits of load shedding, surely, a policy decision - not appropriate for a MUST.

I notice that you've not responded to any of the situations I outlined in the original comment; if you would do so we could perhaps better understand where you're coming from.

To come at it another way, what would you like to see UAs to do when it's discovered that a Log has accepted something which the RFC says it MUST NOT have, perhaps due to a bug?  Is that grounds for distrusting a Log, or is it more likely that it'll be given a pass because it's not actively acting outside the public interest?  
If you feel like it's the latter, then it seems like you should also accept that, in effect, the MUST NOT we're talking about is a SHOULD because otherwise the Log should be immediately distrusted for being non RFC compliant.

AGWA:
@AlCutter wrote:
> To make sure I understand, your position is that it's OK to log arbitrary DER encoded rubbish so long as it's correctly signed, but it's not OK to log actual issued certificates, that are trusted by UAs, and have a slight encoding error in the representation of the signature?

While I'm sympathetic to the argument that if a certificate is trusted by UAs it should be logged, we are moving to a world where in order to be trusted by UAs a certificate must be logged, so logs being strict with their acceptance policy would not be a bad thing and would force CAs to fix encoding bugs.  There can still be logs like Daedalus, Dodo, and Submariner that are not trusted by UAs and can therefore be laxer with what they accept because the security stakes are lower.

> To come at it another way, what would you like to see UAs to do when it's discovered that a Log has accepted something which the RFC says it MUST NOT have, perhaps due to a bug? Is that grounds for distrusting a Log, or is it more likely that it'll be given a pass because it's not actively acting outside the public interest?

Sadly, I think that in such a situation the log needs to be distrusted because it may be an attempt for a log and a CA to collude so the CA can avoid responsibility for misissuance.

AlCutter:
> > To come at it another way, what would you like to see UAs to do when it's discovered that a Log has accepted something which the RFC says it MUST NOT have, perhaps due to a bug? Is that grounds for distrusting a Log, or is it more likely that it'll be given a pass because it's not actively acting outside the public interest?
> Sadly, I think that in such a situation the log needs to be distrusted because it may be an attempt for a log and a CA to collude so the CA can avoid responsibility for misissuance.

If a chain is good enough (this is my "slight sig encoding error" example) to be verified by a UA, then the CA can't really claim that it didn't issue it, right?

Otherwise, if a sig is completely borked, then sure - we have no proof it was issued by them, but the UAs aren't going to accept it either so it's just some bytes which were logged. 

A cert being present in a CT Log shouldn't be construed to confer any special blessing on the correctness of a cert or its issuance - they're just logging things which are sent their way, and, as the RFC paragraph in question explicitly says, the _MUST NOT_ is purely intended as overload protection of the Log, not its correct operation.  

I disagree with your premise, but if the WG consensus agrees with you then a much larger change is going to be required...

AGWA:
> If a chain is good enough (this is my "slight sig encoding error" example) to be verified by a UA, then the CA can't really claim that it didn't issue it, right?

Correct, but now we have to get into what kinds of errors are OK for a log to accept, and 
which aren't, and that's a much larger change than replacing "MUST" with "SHOULD".  Is it worth it?  The easiest and safest option is to just leave it as a "MUST".

> A cert being present in a CT Log shouldn't be construed to confer any special blessing on the correctness of a cert or its issuance - they're just logging things which are sent their way, and, as the RFC paragraph in question explicitly says, the MUST NOT is purely intended as overload protection of the Log, not its correct operation.

The log isn't conferring any blessing on the correctness of the certificate or its issuance.  It's just that the log has to store a valid chain so we know which CA to blame if a logged certificate is misissued.

This isn't a new idea.  Section 4.3 already makes it clear that the requirement for a valid chain is not just to avoid overload:

> If a submission is accepted and an SCT issued, the accepting log MUST store the entire chain used for verification.  This chain MUST include the certificate or precertificate itself, the zero or more intermediate CA certificates provided by the submitter, and the trust anchor used to verify the chain (even if it was omitted from the submission).  The log MUST present this chain for auditing upon request (see Section 5.6).  This prevents the CA from avoiding blame by logging a partial or empty chain

AlCutter:
> > If a chain is good enough (this is my "slight sig encoding error" example) to be verified by a UA, then the CA can't really claim that it didn't issue it, right?
> Correct, but now we have to get into what kinds of errors are OK for a log to accept, and
which aren't, and that's a much larger change than replacing "MUST" with "SHOULD". Is it worth it? The easiest and safest option is to just leave it as a "MUST".

The point of making it a _SHOULD_ is that you don't need to try to enumerate all the ways, many  outside your control, in which you might get bitten, and, for the same reason, it's incredibly dangerous to leave this as a _MUST_ (even if you do try to add a list of exceptions) because if you claim that a Log must be distrusted when it encounters a bug in one of the underlying libraries' and admits a thing it shouldn't have, then a day _will_ come when such a bug is hit and we will lose a significant fraction of the ecosystem in one fell swoop.

The bugs are real enough; without having had to search particularly hard, here are some examples of the sorts of things which could cause this: [one](https://bugzilla.redhat.com/show_bug.cgi?id=1413911), [two](https://www.symantec.com/connect/blogs/common-rsa-implementation-mistake-explained), [three](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-2110).

I'm not trying to pick on OpenSSL; BoringSSL, NSS, SChannel, BouncyCastle, etc. all have had their own share of CVEs, and *anything* with parsing errors, buffer overflows etc. also fall into the same bucket.

Now, Section 5.6 says the log MUST return the chain+anchor which were used to submit the entry, so a Log doesn't have the option of not storing and returning _something_, and so these events are detectable, and can be dealt with in the same way as other policy violations would be handled: a considered response based on evidence and facts.
If a Log is admitting tonnes of garbage, or not returning chains, it's probably not operating in the public interest. If, on the other hand, there are a handful of entries with signatures which are technically broken but can still be verified none-the-less (I'm looking at you, Mr. Non-Minimally-Encoded-Integer), and we know that using _ReallyQuiteInterestingSSL_ is a plausible reason for this, then the ecosystem can say with some confidence that it's likely there was no malfeasance ("and, BTW, please update your software if you haven't already").

I'll also say that CT is, and has always been, about visibility into issuance, not preventing misissuance (which for some value arguably includes badly encoded certs).

I do dig your vision where all the things are _Right and Proper_, and I hope one day it'll come, but the way to get there isn't via a flag day when we lose half the CAs because they can no-longer Log anything their issuing software spits out; who's going to be willing to deploy or rely on infrastructure which is known a priori to either break half the internet, or essentially to be guaranteed to be, or quickly become, non-compliant with its own spec?

AlCutter:
Ooo, BTW @robstradling, having just actually looked at the PR diff (*cough*) I think I'd be ok with just the first MUST -> SHOULD.
I feel like the path building stuff is fine to leave as a MUST NOT as I think we probably do want to require the submitter to present a full chain they consider to be valid for submission, even if it's just something they cooked up with a path finding tool.

AGWA:
>
> The bugs are real enough; without having had to search particularly hard, here are some examples of the sorts of things which could cause this: one, two, three.
>
> I'm not trying to pick on OpenSSL; BoringSSL, NSS, SChannel, BouncyCastle, etc. all have had their own share of CVEs, and anything with parsing errors, buffer overflows etc. also fall into the same bucket.
>
> Now, Section 5.6 says the log MUST return the chain+anchor which were used to submit the entry, so a Log doesn't have the option of not storing and returning something, and so these events are detectable, and can be dealt with in the same way as other policy violations would be handled: a considered response based on evidence and facts.
> If a Log is admitting tonnes of garbage, or not returning chains, it's probably not operating in the public interest. If, on the other hand, there are a handful of entries with signatures which are technically broken but can still be verified none-the-less (I'm looking at you, Mr. Non-Minimally-Encoded-Integer), and we know that using ReallyQuiteInterestingSSL is a plausible reason for this, then the ecosystem can say with some confidence that it's likely there was no malfeasance ("and, BTW, please update your software if you haven't already").

I completely agree that if a log can show that a broken signature was due to a bug and not malfeasance then it should not necessarily be distrusted.  Leaving the requirement a MUST would not preclude that - several logs have been left trusted in Chrome despite violating MUSTs in RFC6962, or other hard policy requirements such as 99% uptime.

What I oppose is weakening the security properties of Certificate Transparency by changing this MUST to a SHOULD.  The RFC should describe how a correct, bug-free implementation can provide a verifiable Certificate Transparency log.  Policy can decide how to deal with logs that violate the requirement.

> I'll also say that CT is, and has always been, about visibility into issuance, not preventing misissuance (which for some value arguably includes badly encoded certs).

Visibility into issuance is meaningless if you don't know who is doing the issuing, which is why ensuring a valid chain is so important.

I would also say that CT has always been about not having to blindly trust logs.  If this weren't the case, it would just be a database of certificates and not a Merkle Tree.  Unfortunately, some log bugs are indistinguishable from log malfeasance.  Are we really going to give up this property of CT because chain validation is hard?

> I do dig your vision where all the things are Right and Proper, and I hope one day it'll come, but the way to get there isn't via a flag day when we lose half the CAs because they can no-longer Log anything their issuing software spits out; who's going to be willing to deploy or rely on infrastructure which is known a priori to either break half the internet, or essentially to be guaranteed to be, or quickly become, non-compliant with its own spec?

Is this really going to break half the Internet?  RFC6962 has been deployed despite this requirement being a MUST, and Chrome has successfully required CT for EV treatment since 2015, and all certificates from the world's largest CA since 2016.  Surely if this requirement were causing serious problems we would have noticed by now.

robstradling:
@AlCutter I've changed the 2nd and 3rd SHOULD NOT back to MUST NOT.

ekr:
First, let me say that this is a WG decision. I raised it in my AD review because it struck me as odd.

With that out of the way, I don't understand the argument that @AGWA is making:

> This requirement is not just for anti-spam, but for security. Logs need to contain cryptographic proof of which CA misissued a certificate. If you allow logs to include invalid chains, then a CA could misissue a certificate, collude with a log to log it with an invalid chain, and when someone notices the misissued certificate, the CA can plausibly claim it never issued it.

Why is this a useful form of attack? Imagine that I just encountered a certificate which was correctly signed (i.e., the signature was correct) but the chain was somehow bogus, how would that change the situation with respect to being able to hold the CA to account?

It seems to me that if there is a security requirement here, it is that clients should not consider certificates to have been logged (and then hence potentially invalid) unless the log entry contains a valid chain.

AlCutter:
> I completely agree that if a log can show that a broken signature was due to a bug and not malfeasance then it should not necessarily be distrusted.

That's not what you said earlier:

>> [al] To come at it another way, what would you like to see UAs to do when it's discovered that a Log has accepted something which the RFC says it MUST NOT have, perhaps due to a bug? Is that  grounds for distrusting a Log, or is it more likely that it'll be given a pass because it's not actively  acting outside the public interest?

> [agwa] Sadly, I think that in such a situation the log needs to be distrusted because it may be an attempt for a log and a CA to collude so the CA can avoid responsibility for misissuance.

:)

> What I oppose is weakening the security properties of Certificate Transparency by changing this MUST to a SHOULD.

I don't see the weakening of security properties, maybe I'm missing something, but you've mentioned this multiple times and talked about a bad Log & CA colluding to log things which shouldn't have been, so maybe we can talk a bit more about it.

AIUI you're worried about a CA logging a thing which nobody can attribute to them, and this confers legitimacy on the cert because it has SCTs, and so a UA will trust the cert when later presented with the full chain which is owned only by the attacker (let's say it's issued via a sekret intermediate so we can't just try to verify the EE sig with keys from all known CA:true certs)?

There are two obvious ways a Log can collude with a bad CA here:
a) it issues the SCT but does not integrate the cert (this is with the MUST left as-is)
b) it simply drops the chain, issues the SCT, and integrates the cert (this, I think, is your concern)

I would say that these are essentially identical in outcome:
a) once detected by lack of inclusion proof, the Log gets distrusted in ~24h due to policy violation ("correct operation of the log according to the RFC")
b) once the entry with no chain is detected by a Monitor, the finding is reported and Log gets distrusted due to policy ("correct operation - 5.6 MUST is failed", and/or "not operating in the public interest")

No "real" Log is going to choose either (a) or (b) willingly, because it gets burnt, and if it's a Bad Log, then it'll do either (a) or (b) regardless of any RFC MUST. 
(b) is arguably better for the bad CA since it's harder to figure out whodunit, but there are only so many times they can pull this stunt - it's pretty expensive and time-consuming to stand up logs, get them trusted by CAs & UA, only to burn them after a single use.  
There's obviously also a non-zero risk that the sekret intermediate will out (perhaps by being logged itself), and then the game's well and truly up.

It's late here, what am I missing?

> CT has always been about not having to blindly trust logs. If this weren't the case, it would just be a database of certificates and not a Merkle Tree. Unfortunately, some log bugs are indistinguishable from log malfeasance. Are we really going to give up this property of CT because chain validation is hard?

I don't see how this affects trust in the Logs?
The verifiable properties of Logs does not change in any way due to this change, does it?

> Policy can decide how to deal with logs that violate the requirement.

This sounds rather like a SHOULD NOT, then doesn't it? 
i.e. it's not a technical requirement for conformance with the spec., but rather a thing which is ok to do under some particular circumstances, or in RFC2119 language:
> [RFC2119] **SHOULD**   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item.

No?

> Is this really going to break half the Internet? RFC6962 has been deployed despite this requirement being a MUST, and Chrome has successfully required CT for EV treatment since 2015, and all certificates from the world's largest CA since 2016. Surely if this requirement were causing serious problems we would have noticed by now.

I'll offer another interpretation: nobody is verifying the strict adherence to this MUST, because there's not really any way to do so since all the tools use the same permissive verification libraries, which, I argue, makes it a tacit SHOULD :)

AGWA:
@ekr wrote:

> Why is this a useful form of attack? Imagine that I just encountered a certificate which was correctly signed (i.e., the signature was correct) but the chain was somehow bogus, how would that change the situation with respect to being able to hold the CA to account?

With a valid signature but bogus chain, you would know only the intermediate CA which issued the certificate, but not the root which signed the intermediate.  So you wouldn't know which CA should be pulled from root programs.

> It seems to me that if there is a security requirement here, it is that clients should not consider certificates to have been logged (and then hence potentially invalid) unless the log entry contains a valid chain.

Ideally, but it's not clear how this can be done.  It would be possible if the Merkle leaf contained the whole chain, not just the TBSCertificate, but clients might build a different chain from the one that is logged, so reconstructing the entry to verify the SCT signature would fail.  Instead, I think this behavior has to be enforced by monitors.

AGWA:
@AlCutter wrote:

>> I completely agree that if a log can show that a broken signature was due to a bug and not malfeasance then it should not necessarily be distrusted.
>
> That's not what you said earlier:
>>> [al] To come at it another way, what would you like to see UAs to do when it's discovered that a Log has accepted something which the RFC says it MUST NOT have, perhaps due to a bug? Is that grounds for distrusting a Log, or is it more likely that it'll be given a pass because it's not actively acting outside the public interest?
>>
>>  [agwa] Sadly, I think that in such a situation the log needs to be distrusted because it may be an attempt for a log and a CA to collude so the CA can avoid responsibility for misissuance.

To be clear, I don't think the log needs to be distrusted *if it can show that it was a bug and not malfeasance* (e.g. a signature that is slightly misencoded but still validates). In the default case of accepting a bad chain, I maintain that log distrust is necessary.

>>  What I oppose is weakening the security properties of Certificate Transparency by changing this MUST to a SHOULD.
>
> I don't see the weakening of security properties, maybe I'm missing something, but you've mentioned this multiple times and talked about a bad Log & CA colluding to log things which shouldn't have been, so maybe we can talk a bit more about it.
>
> AIUI you're worried about a CA logging a thing which nobody can attribute to them, and this confers legitimacy on the cert because it has SCTs, and so a UA will trust the cert when later presented with the full chain which is owned only by the attacker (let's say it's issued via a sekret intermediate so we can't just try to verify the EE sig with keys from all known CA:true certs)?
>
> There are two obvious ways a Log can collude with a bad CA here:
> a) it issues the SCT but does not integrate the cert (this is with the MUST left as-is)
> b) it simply drops the chain, issues the SCT, and integrates the cert (this, I think, is your concern)
>
> I would say that these are essentially identical in outcome:
> a) once detected by lack of inclusion proof, the Log gets distrusted in ~24h due to policy violation ("correct operation of the log according to the RFC")
> b) once the entry with no chain is detected by a Monitor, the finding is reported and Log gets distrusted due to policy ("correct operation - 5.6 MUST is failed", and/or "not operating in the public interest")
>
> No "real" Log is going to choose either (a) or (b) willingly, because it gets burnt, and if it's a Bad Log, then it'll do either (a) or (b) regardless of any RFC MUST.
> (b) is arguably better for the bad CA since it's harder to figure out whodunit, but there are only so many times they can pull this stunt - it's pretty expensive and time-consuming to stand up logs, get them trusted by CAs & UA, only to burn them after a single use.
> There's obviously also a non-zero risk that the sekret intermediate will out (perhaps by being logged itself), and then the game's well and truly up.
>
> It's late here, what am I missing?

Scenario (b) captures my concerns, although I would broaden it to include invalid chains, not just missing chains.  It sounds like we agree that if (b) happens, the log needs to be distrusted - am I correct?  Where we disagree is that I don't think 5.6 prohibits (b) from happening.  It requires the log to return the chain used for submission, so if we remove the requirement that submitted chains be valid, we effectively remove the requirement that the log produce a valid chain upon demand, as the log can plausibly claim that the certificate was submitted without one.

(Also, we shouldn't rely on Chrome's "operating in the public interest" requirement, because it's not part of the document, and other UAs who adopt CT might not know the need for this additional requirement.)

>>  CT has always been about not having to blindly trust logs. If this weren't the case, it would just be a database of certificates and not a Merkle Tree. Unfortunately, some log bugs are indistinguishable from log malfeasance. Are we really going to give up this property of CT because chain validation is hard?
>
> I don't see how this affects trust in the Logs?
> The verifiable properties of Logs does not change in any way due to this change, does it?

One of the crucial verifiable properties is that a log can provide a valid chain from any entry to an accepted trust anchor.  I'm worried about weakening CT so a log only has to *try* to satisfy this property, and if it fails, it's assumed to be a mistake and not malfeasance - essentially placing blind trust in the log.

>> Policy can decide how to deal with logs that violate the requirement.
>
> This sounds rather like a SHOULD NOT, then doesn't it?
> i.e. it's not a technical requirement for conformance with the spec., but rather a thing which is ok to do under some particular circumstances, or in RFC2119 language:
>
>>      [RFC2119] SHOULD This word, or the adjective "RECOMMENDED", mean that there
>>      may exist valid reasons in particular circumstances to ignore a
>>      particular item.
>
> No?

Given the security implications, I think it's unwise to let the implementer decide what the "valid reasons" are.  I would rather 6962-bis explicitly list any necessary exceptions.  Failing that, leave it a MUST NOT and treat the acceptance of a bad chain as a compliance bug that requires fixing, but not necessarily distrust.

>> Is this really going to break half the Internet? RFC6962 has been deployed despite this requirement being a MUST, and Chrome has successfully required CT for EV treatment since 2015, and all certificates from the world's largest CA since 2016. Surely if this requirement were causing serious problems we would have noticed by now.
>
> I'll offer another interpretation: nobody is verifying the strict adherence to this MUST, because there's not really any way to do so since all the tools use the same permissive verification libraries, which, I argue, makes it a tacit SHOULD :)

This would be sad if true.  Do we have any way of determining how severe a problem this actually is?  It seems sub-optimal to make a decision without data in hand.

eranmes:
I'd like to point out that even if there's no way to verify strict adherence to these MUSTs, they are here to guide the implementation:
* A submission must never be admitted to the log without validation. The validation may be lax (beyond what RFC5280 permits) or just buggy, but it should always take place. I don't see a reason to relax that requirement of logs.
* Using different intermediates to complete a partially-submitted chain just makes auditing difficult: It may not be a security issue, but would make it harder to identify particular intermediate CAs that have mis-issued certificates by potentially obscuring them.

benlaurie:
I am slightly confused by some of this discussion: if the argument is that a signature might be slightly malformed but still verifies correctly in some UA, then why is that not already permitted under the "lax 5280" provision?

ekr:
On Wed, Nov 22, 2017 at 5:51 AM, Ben Laurie <notifications@github.com>
wrote:

> I am slightly confused by some of this discussion: if the argument is that
> a signature might be slightly malformed but still verifies correctly in
> some UA, then why is that not already permitted under the "lax 5280"
> provision?
>

But then I don't even know what the MUST means.

Perhaps someone could give an example of two malformed chains, one of which
the log would be required to reject and one of which it would be allowed to
accept?

AlCutter:
@benlaurie, that's _one_ of the arguments - but you tell me, does "MUST ... valid signature chain" allow for encoding errors?

benlaurie:
Well, I agree with @ekr that I don't really know what the MUST means in this context. Seems to me that MUST -> SHOULD merely clarifies what the situation already is, to the extent the existing situation is intelligible at all.

Objectionable behaviour by logs along the lines of accepting spam or whatever can be dealt with by policy.

I can't see how relaxing this can lead to other attacks.

eranmes:
I've had another go, at @robstradling 's request, in PR #293

robstradling:
Closing in favour of #294.

PR #295, opened 2018-04-20
A-D review: Remove the Preventing Tracking Clients section.

robstradling:
EKR wrote:
`I think "tracked or traced" by who.  As far as a deterministic scheme, that seems like a wg decision, but what would be the implication for RSA-PSS?`

Eran, Al and I discussed this.  Al wrote...
`I don't think this document should be talking about this at all; this seems like a concession to a gossip design which hasn't yet been experimented with, built, or standardised...
Would this section not better live in whatever gossip docs which build upon -bis, where it's important for that model's security/privacy guarantees?`
...and then we agreed to remove this section.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -2113,19 +2113,6 @@ issued too closely together, proving violation of the STH issuance rate limit;
 or an STH with a root hash that does not match the one calculated from a copy of
 the log, proving violation of the append-only property.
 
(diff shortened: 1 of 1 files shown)


AGWA:
This should be discussed on the mailing list before merging to ensure there is WG consensus.

eranmes:
@AGWA do you recall the results of that discussion?

robstradling:
Closing, as per https://mailarchive.ietf.org/arch/msg/trans/I47LvwrebWZMEDOav0zHtJao4PE

PR #297, opened 2018-06-15
Switch intended status of 6962-bis from Standards Track to Experimental.

robstradling:
As per discussion with the TRANS Chairs, Security A-Ds, and the other 6962-bis authors.


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -2,7 +2,7 @@
 title: "Certificate Transparency Version 2.0"
 docname: draft-ietf-trans-rfc6962-bis-29
 obsoletes: 6962
-category: "std"
+category: exp
```
eranmes: No need for quotes?

robstradling: No need for quotes.

PR #299, opened 2018-08-24
Changing NTP time to Unix time

eranmes:
As pointed out on the trans mailing list, the definition of the
timestamp in the document conflated NTP time and Unix time, see:
https://mailarchive.ietf.org/arch/msg/trans/wwQYt2vsKOGIHqfSipJXHd2-okI

Make it clear that Unix time is actually used in the timestamps.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -44,7 +44,6 @@ normative:
   RFC4648:
   RFC5280:
   RFC5652:
-  RFC5905:
   RFC6066:
   RFC6960:
   RFC6961:
@@ -76,6 +75,11 @@ normative:
     author:
       org: NIST
     date: 2013-07-01
+  UNIXTIME:
+    target: http://pubs.opengroup.org/onlinepubs/9699919799.2016edition/basedefs/V1_chap04.html#tag_04_16
+    title: The Open Group Base Specifications Issue 7 IEEE Std 1003.1-2008, 2016 Edition
+    author:
+      org: IEEE
 
 informative:
   RFC6234:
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -914,10 +918,9 @@ log's parameters.
     } TimestampedCertificateEntryDataV2;
 ~~~~~~~~~~~
 
-`timestamp` is the NTP Time [RFC5905] at which the certificate or precertificate
-was accepted by the log, measured in milliseconds since the epoch (January 1,
-1970, 00:00 UTC), ignoring leap seconds. Note that the leaves of a log's Merkle
-Tree are not required to be in strict chronological order.
+`timestamp` is unsigned Unix Time [UNIXTIME] at which the certificate or precertificate
```
robstradling: Unix Time is 32-bit and measured in seconds, whereas we're using 64-bit values that are measured in milliseconds.  I think it makes sense to reference the POSIX document though, because we're using the same Epoch, the same byte order, and we treat leap seconds the same way.

My proposed text:
"timestamp" is the date and time at which the certificate or precertificate was accepted by the log, in the form of a 64-bit unsigned number of milliseconds elapsed since the Unix Epoch (1 January 1970 00:00:00 UTC - see [UNIXTIME]), ignoring leap seconds, in network byte order. Note that the leaves of a log's Merkle Tree are not required to be in strict chronological order.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -985,8 +988,7 @@ The log stores information about its Merkle Tree in a `TreeHeadDataV2`:
 
 The length of NodeHash MUST match HASH_SIZE of the log.
 
-`timestamp` is the current NTP Time [RFC5905], measured in milliseconds since
-the epoch (January 1, 1970, 00:00 UTC), ignoring leap seconds.
+`timestamp` is the current, unsigned Unix Time [UNIXTIME].
```
robstradling: My proposed text:
"timestamp" is the current date and time, in the form of a 64-bit unsigned number of milliseconds elapsed since the Unix Epoch (1 January 1970 00:00:00 UTC - see [UNIXTIME]), ignoring leap seconds, in network byte order.

draft-ietf-trans-rfc6962-bis.md
eranmes: Thanks for highlighting that, accepted your text.
I'm wondering if "accepted by the log" is ambiguous and would be subject to language lawyering. I think not (because there's already a marked difference between acceptance of a submission and incorporation of said submission into the tree).

draft-ietf-trans-rfc6962-bis.md
eranmes: Changed.

eranmes:
@robstradling PTAL

draft-ietf-trans-rfc6962-bis.md
robstradling: I agree that it's not ambiguous.

PR #300, opened 2018-08-24
Output errors in compliance with RFC7807

eranmes:
Change the error output format to be the Problem Details Object from
RFC7807.

Seems pretty straightforward given that the 6962-bis errors do not
have unique requirements and so can comply with said RFC.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -50,6 +50,7 @@ normative:
   RFC7159:
   RFC7231:
   RFC7633:
+  RFC7807:
   RFC7924:
   RFC8032:
   RFC8446:
@@ -1153,26 +1154,29 @@ requirement above).
 
 If the log is unable to process a client's request, it MUST return an HTTP
 response code of 4xx/5xx (see [RFC7231]), and, in place of the responses
-outlined in the subsections below, the body SHOULD be a JSON structure
-containing at least the following field:
+outlined in the subsections below, the body SHOULD be a JSON Problem Details
+Object (see [RFC7807] Section 3), containing:
 
-error_message:
+type:
+: A URN reference identifying the problem. To facilitate automated response
+  to errors, this document defines a set of standard tokens for use in the
+  `type` field, within the URN namespace of: "urn:ietf:params:trans:error:".
+  Other than the "malformed" error type defined in this section, each error
+  type is specific to the type of request. Specific errors are defined in the
+  respective sections below.
+
+detail:
 : A human-readable string describing the error which prevented the log from
   processing the request.
 
 : In the case of a malformed request, the string SHOULD provide sufficient
   detail for the error to be rectified.
 
-error_code:
-: An error code readable by the client. Other than the generic codes detailed
-  here, each error code is specific to the type of request. Specific
-  errors are specified in the respective sections below. Error codes are fixed
-  text strings.
 
 |---------------+---------------------------------------------|
-| Error Code    | Meaning                                     |
+| type          | detail                                      |
 |---------------+---------------------------------------------|
-| not compliant | The request is not compliant with this RFC. |
+| malformed     | The request is not compliant with this RFC. |
 |---------------+---------------------------------------------|
 
 e.g., In response to a request of `/ct/v2/get-entries?start=100&end=99`, the log
@@ -1181,8 +1185,8 @@ following:
 
 ~~~~~~~~~~~
     {
-        "error_message": "'start' cannot be greater than 'end'",
-        "error_code": "not compliant",
+        "type": "urn:ietf:params:trans:error:endBeforeStart",
+        "detail": "'start' cannot be greater than 'end'"
     }
 ~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1179,8 +1185,9 @@ following:
 
 ~~~~~~~~~~~
     {
-        "error_message": "'start' cannot be greater than 'end'",
-        "error_code": "not compliant",
+        "type": "https://example.net/get-entries-error"
+        "title": "not compliant",
+        "detail": "'start' cannot be greater than 'end'",
```
robstradling: Using URNs (a la ACME), "not compliant" would become a camelCase suffix to the "type" parameter.  i.e.,

```
{
    "type": "urn:ietf:params:ct:error:notCompliant",
    "detail": "'start' cannot be greater than 'end'"
}
```

A few other thoughts...
"notCompliant" seems unnecessarily generic.  For problems with request parameters, ACME uses "malformed" and RFC7807 suggests "validation-error".
RFC7807 has the concept of subproblems, permitting details of multiple similar problems to be conveyed.  ACME does this.  Using https://tools.ietf.org/html/draft-ietf-acme-acme-14#section-6.6.1 and https://tools.ietf.org/html/rfc7807#section-3 for inspiration, this example might look something like this...

```
{
    "type": "urn:ietf:params:ct:error:malformed",
    "detail": "The request message was malformed",
    "subproblems": [
        {
            "name": "start",
            "detail": "must not be greater than 'end'"
        },
        {
            "name": "end",
            "detail": "must not be less than 'start'"
        }
    ]
}
```

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1151,24 +1152,29 @@ requirement above).
...
 
-error_message:
+type:
+: A URI reference identifying the problem. For example, a reference to this
+  document.
+
+title:
+: An error code readable by the client. Other than the generic codes detailed
+  here, each error code is specific to the type of request. Specific
+  errors are specified in the respective sections below. Error codes are fixed
```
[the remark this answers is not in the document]
eranmes: Done.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1179,8 +1185,9 @@ following:
 
 ~~~~~~~~~~~
     {
-        "error_message": "'start' cannot be greater than 'end'",
-        "error_code": "not compliant",
+        "type": "https://example.net/get-entries-error"
```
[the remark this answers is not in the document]
eranmes: Done.

draft-ietf-trans-rfc6962-bis.md
eranmes: (1) Switched to 'malformed' as suggested.
(2) Not sure the 'subproblems' structure makes a lot of sense: All but one of the errors would fall into the 'malformed' category (the exception being 'shutdown' which is a bad log state).

eranmes:
@robstradling PTAL.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1147,26 +1148,29 @@ requirement above).
...
 
-error_code:
-: An error code readable by the client. Other than the generic codes detailed
-  here, each error code is specific to the type of request. Specific
-  errors are specified in the respective sections below. Error codes are fixed
-  text strings.
 
 |---------------+---------------------------------------------|
-| Error Code    | Meaning                                     |
+| Type          | Detail                                      |
```
robstradling: I think "type" and "detail" (i.e., no capitals) would be less confusing column headers for these tables.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1147,26 +1148,29 @@ requirement above).
...
-
-error_message:
+outlined in the subsections below, the body SHOULD be a JSON Problem Details
+Object (see [RFC7807] Section 3), containing:
+
+type:
+: A URN reference identifying the problem. To facilitate automated diagnostics
+  of errors, this document defines a set of standard tokens for use in the
+  `type` field, within the URN namespace of: "urn:ietf:params:trans:error:".
+  Other than the generic codes detailed here, each error type is specific to
```
robstradling: We're only specifying one generic error: "malformed".

I suggest: s/the generic codes detailed here/the generic "malformed" error type defined in this section/

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1147,26 +1148,29 @@ requirement above).
...
-  here, each error code is specific to the type of request. Specific
-  errors are specified in the respective sections below. Error codes are fixed
-  text strings.
 
 |---------------+---------------------------------------------|
-| Error Code    | Meaning                                     |
+| Type          | Detail                                      |
 |---------------+---------------------------------------------|
-| not compliant | The request is not compliant with this RFC. |
+| malformed     | The request is not compliant with this RFC. |
```
robstradling: This table seems to be saying that "detail" will be always be the fixed string "The request is not compliant with this RFC", which obviously conflicts with the previous sentence (i.e., "...SHOULD provide sufficient detail for the error to be rectified").

How about changing "The request is not compliant with this RFC" to "\<sufficient detail for the error to be rectified>"?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1147,26 +1148,29 @@ requirement above).
...
 response code of 4xx/5xx (see [RFC7231]), and, in place of the responses
-outlined in the subsections below, the body SHOULD be a JSON structure
-containing at least the following field:
-
-error_message:
+outlined in the subsections below, the body SHOULD be a JSON Problem Details
+Object (see [RFC7807] Section 3), containing:
+
+type:
+: A URN reference identifying the problem. To facilitate automated diagnostics
```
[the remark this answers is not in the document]
eranmes: Sounds reasonable, done.

draft-ietf-trans-rfc6962-bis.md
eranmes: Done.

draft-ietf-trans-rfc6962-bis.md
eranmes: Done, here and throughout.

draft-ietf-trans-rfc6962-bis.md
eranmes: Per OOB discussion:
* Left 'malformed' as a general error.
* Specified type and detail for each of the cases where 'malformed' might have been used.

eranmes:
@robstradling PTAL.

robstradling:
Thanks @eranmes.  I'm going to merge this and then propose a follow-on PR to clean up a couple of things.

robstradling:
My follow-on PR is #301.

PR #305, opened 2018-12-27
A-D Review: Clarify submission acceptance rules.

robstradling:
This PR proposes to split the "Accepting Submissions" section into two subsections, in order to clarify what are the minimum rules and what's left to the log's discretion.

It also enumerates the minimum rules.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -767,12 +767,12 @@ If the acceptance criteria are met, the log SHOULD accept the submission. (A log
 may decide, for example, to temporarily reject acceptable submissions to protect
 itself against denial-of-service attacks).
 
-### Minimum Acceptance Criteria {#minimum_criteria}
-
 The log SHALL allow retrieval of its list of accepted trust anchors (see
 {{get-anchors}}), each of which is a root or intermediate CA certificate. This
 list might usefully be the union of root certificates trusted by major browser
 vendors.
+
+### Minimum Acceptance Criteria {#minimum_criteria}
 
 To ensure that logged certificates and precertificates are attributable to an
 accepted trust anchor, and to set clear expectations for what monitors would
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -779,11 +779,10 @@ accepted trust anchor, and to set clear expectations for what monitors would
 find in the log, and to avoid being overloaded by invalid submissions, the log
 MUST reject a submission if any of the following conditions are not met:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -757,39 +757,61 @@ Final STH:
...
 vendors.
 
+To ensure that logged certificates and precertificates are attributable to an
+accepted trust anchor, and to set clear expectations for what monitors would
+find in the log, and to avoid being overloaded by invalid submissions, the log
+MUST reject a submission if any of the following conditions are not met:
+
+* The submitted certificate or precertificate, together with the zero or more
+  intermediate CA certificates provided in the submission (in the order
+  provided), MUST form a valid signature chain to an accepted trust anchor. (The
```
ekr: Does "valid" here just mean "the signatures verify" or something else?

robstradling: Yes, "valid signature chain" is intended to just mean "the signatures verify".

I think what you're implying is that the term "valid" is probably overloaded in people's minds, and I guess I'd agree.  Shall I rework this to say something like "each cert's signature must be verifiable by the public key of the next cert in the chain"?

ekr: That seems like a good idea.

robstradling: Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/9665c108c20c8a1ee755ffd3d22401c343f152db.

@ekr Does that look OK?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -757,39 +757,60 @@ Final STH:
...
+against the minimum acceptance criteria (see {{minimum_criteria}}) and against
+the log's discretionary acceptance criteria (see {{discretionary_criteria}}).
+
+If the acceptance criteria are met, the log SHOULD accept the submission. (A log
+may decide, for example, to temporarily reject acceptable submissions to protect
+itself against denial-of-service attacks).
+
+### Minimum Acceptance Criteria {#minimum_criteria}
 
 The log SHALL allow retrieval of its list of accepted trust anchors (see
```
eranmes: Nit: It seems like this section belongs at the end / outside the minimum acceptance criteria section? It seems out of place at the beginning of this  section.

robstradling: Good point.  Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/e3f549fbf5149092a15c24808e1352ed512b0530

PR #306, opened 2018-12-28
A-D Review: Be more explicit about what a TLS client does after validating SCTs.

robstradling:


ekr:
But then what's the normative content of the requirement to validate all the SCTs? It appears to have no externally visible properties. Here's the relevant guidance from 2119:

```
   Imperatives of the type defined in this memo must be used with care
   and sparingly.  In particular, they MUST only be used where it is
   actually required for interoperation or to limit behavior which has
   potential for causing harm (e.g., limiting retransmisssions)  For
```

This gets violated a lot, so I think it's an open question how dispositive it is, but I think it's a good question to ask.

robstradling:
@ekr The WG decided some time ago that 6962-bis should describe the CT Protocol but not attempt to dictate CT Policy.  This means that 6962-bis needs to document the common technical building blocks (e.g., SCT validation, inclusion proof validation) that "CT-using TLS clients" need to use to evaluate compliance.  Client policies that use those common technical building blocks to determine (non-)compliance, and client behaviour that results from that determination of (non-)compliance, are both out of scope for this document.

I don't know what else to say, other than...please propose text!

ekr:
I propose that you remove the text:

"  In addition to normal validation of the server certificate and its
   chain, CT-using TLS clients MUST validate each received SCT for which
   they have the corresponding log's parameters. "

and replace it with:

"In order to make use of an SCT, clients MUST first validate it."

robstradling:
@ekr In https://github.com/robstradling/certificate-transparency-rfcs/commit/d8c282d753bb7f56c9ac8ae3458543de270b2367 I've incorporated your proposed text and reworked the "Validating SCTs" section a bit.  PTAL.

ekr:
LGTM

PR #309, opened 2019-06-17
IESG Last Call: Address Alissa Cooper's DISCUSS and COMMENT feedback

robstradling:
> On 13th March 2019, Alissa Cooper wrote:
> Glad to see this revision of the protocol. My comments and questions should be easy to address.
> 
> = Section 10.2, 10.4, 10.5 =
> 
> A Specification Required registry policy implies expert review. So a registry policy of "Specification Required and Expert Review" is duplicative; it should just say "Specification Required." I know this seems trivial but there has been so much confusion about this through the years that it is important to be precise.

Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/0a21e8c769bc8d5c0d28bb4c610175476187478a

> = Section 10.3 =
> 
> This section needs to state what the registry policy is for the code points not already registered (presumably Expert Review given 10.3.1, but it needs to be explicit).

Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/7cd3471548c903fd891a99227bf081ca51939470

> = Section 10.6.1 =
> 
> Using the term "Parameters Required" as a capitalized term is confusing. FCFS registries by definition can require additional information to be provided in order to get something registered. For avoidance of confusion I think the assignment policy should be listed as First Come First Served and the requirement that parameters be included in the application can use a normative MUST in the last paragraph if there is concern that the parameters won't be supplied.

Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/704a71a18457b4558ce26fe4be519d6ea06a729a

> However, I also wonder what will be done with the parameters that are supplied. Is IANA expected to just maintain them privately, or to publish them?

On reflection, I think it makes little sense for IANA to handle most of these parameters, since some of them (e.g., Maximum Chain Length, Maximum Merge Delay, STH Frequency Count, Final STH) may change over time for operational reasons that really needn't concern IANA.  Therefore, this PR changes this registry to contain just the Log's Base URL, along with (as advised by https://tools.ietf.org/html/rfc8126#section-9.5) an Owner and Contact (from whom the other parameter values may be obtained).

> What is expected to appear in the 'Log' column in the registry?

Clarified as "Log's Base URL".

> In Section 1.1, please use the RFC 8174 boilerplate in lieu of the RFC 2119 boilerplate.

Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/efb1229cf290f88bbb80016268f77d2fd8193cca

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -50,6 +50,7 @@ normative:
   RFC7633:
   RFC7807:
   RFC8032:
+  RFC8174:
   RFC8259:
   RFC8446:
   HTML401:
@@ -195,8 +196,9 @@ such mechanisms are outside the scope of this document.
 ## Requirements Language
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
-"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in [RFC2119].
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
+when, and only when, they appear in all capitals, as shown here.
 
 ## Data Structures {#data_structures}
 
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1958,14 +1958,18 @@ the log.
 IANA is asked to establish a registry of signature algorithm values, named
 "CT Signature Algorithms", that initially consists of:
 
-|--------------------------------+----------------------------------------------------+--------------------------------|
-| SignatureScheme Value          | Signature Algorithm                                | Reference / Assignment Policy  |
-|--------------------------------+----------------------------------------------------+--------------------------------|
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -195,8 +196,9 @@ such mechanisms are outside the scope of this document.
 ## Requirements Language
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
-"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in [RFC2119].
+"SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
```
robstradling: @eranmes I'm afraid I couldn't figure out how to express the RFC8174 boilerplate without resorting to using implicit references.  (I know we'd been avoiding using implicit references in this document, but I honestly don't remember why ;-) ).

eranmes: Not sure either. LGTM (but I'll point out we  don't use the 'NOT RECOMMENDED' term in the document)

robstradling: Thanks Eran.  I agree that we don't use "NOT RECOMMENDED".  However, Alissa asked us to "use the RFC 8174 boilerplate", which I presume means "copy and paste the RFC 8174 boilerplate, without editing it".

PR #311, opened 2019-07-04
IESG Last Call: Satisfy BCP 190 by requiring each endpoint to be specified as a log parameter

robstradling:
Using a well-known suffix for log endpoint URLs did not go down well with the TRANS WG, but there seems to be consensus for satisfying BCP 190 using a different approach: by requiring each endpoint to be specified as a log parameter.

With this approach, the "Base URL" field is no longer needed.  "Base URL" was also being used as a human-readable description of each entry in the Log ID Registry, so this PR also introduces a "Log Name" field to use for that purpose.


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1221,7 +1236,7 @@ minimum time for the client to wait before retrying the request.
 
 ## Submit Entry to Log {#submit-entry}
 
-POST https://{host}/.well-known/ct/v2/{prefix}/submit-entry
+POST {`submit-entry URL`}
```
eranmes: Nit: It would be nice to add an example URL (e.g. https://log.example.com/ct/v2/submit-entry) so we're more likely to end up with a unified log url organization.

robstradling: I didn't include any examples because I was taking "under-specifying it right now is the only option" (to quote your message on the mailing list) to what I thought was its logical conclusion.  ;-)

eranmes:
Good point. Let's leave it as-is.

draft-ietf-trans-rfc6962-bis.md
robstradling: If we were to include an example such as https://log.example.com/ct/v2/submit-entry, then what are the chances that some implementers will lazily assume that every log uses "/ct/v2/submit-entry" rather than make the effort of obtaining the actual URL from the log's parameters?

robstradling: I think omitting examples is actually preferable in this case.

robstradling:
> The log parameters are not URLs, but URL templates.
> The variables that can appear in the templates need to be defined as well. That is, 'first', 'second', 'hash', 'start, and 'end' for various templates.
> Otherwise the spec is still forcing URL structure on servers (ie variables MUST be querystring fields with these given names).

@manger Thanks for your comment.  I think we should discuss this on the TRANS list before making further changes to this PR.

robstradling:
Closing this PR, per agreement on the TRANS list.

PR #315, opened 2019-10-03
Clarify some implications of the binding intent to issue.

robstradling:
Issuing a precertificate signals the CA's binding intent to issue the corresponding certificate.  Experience with CTv1 and recent discussions on the TRANS and mozilla.dev.security.policy lists have shown that some of the implications of this are not obvious, and so this PR attempts to clarify some of these implications.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -680,12 +680,12 @@ construct a valid certificate.
 
 ### Binding Intent to Issue  {#binding_intent_to_issue}
 
-Under normal circumstances, there will be a short time period between
-precertificate submission and issuance of the corresponding certificate.
-Occasional delays are to be expected though (e.g., due to log server downtime),
-and in some cases the CA might not actually issue the corresponding certificate.
-Nevertheless, a precertificate's `signature` indicates the CA's binding intent
-to issue the corresponding certificate, which means that:
+Under normal circumstances, there will be a short delay between precertificate
+submission and issuance of the corresponding certificate. Longer delays are to
+be expected occasionally (e.g., due to log server downtime), and in some cases
+the CA might not actually issue the corresponding certificate. Nevertheless, a
+precertificate's `signature` indicates the CA's binding intent to issue the
+corresponding certificate, which means that:
 
 * Misissuance of a precertificate is considered equivalent to misissuance of
   the corresponding certificate. The CA should expect to be held to account,
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -668,11 +668,8 @@ following profile:
     * A message-digest attribute whose value is the message digest of
       `SignedData.encapContentInfo.eContent`.
   * `signatureAlgorithm` MUST be the same OID as `TBSCertificate.signature`.
-  * `signature` MUST be from the same (root or intermediate) CA that will
-    ultimately issue the certificate. This signature indicates the CA's intent
-    to issue the certificate. This intent is considered binding (i.e.,
-    misissuance of the precertificate is considered equivalent to misissuance of
-    the corresponding certificate).
+  * `signature` MUST be from the same (root or intermediate) CA that intends to
+    issue the corresponding certificate (see {{binding_intent_to_issue}}).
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -681,6 +678,30 @@ value will not be a valid X.509v3 signature that could be used in conjunction
 with the TBSCertificate (from `SignedData.encapContentInfo.eContent`) to
 construct a valid certificate.
 
+### Binding Intent to Issue  {#binding_intent_to_issue}
+
+Under normal circumstances, there will be a short time period between
+precertificate submission and issuance of the corresponding certificate.
+Occasional delays are to be expected though (e.g., due to log server downtime),
```
eranmes: Nit "Longer delays" ? It seems to me this case is distinct from the "short time period between precertificate submission and issuance of the corresponding certificate" so may be worth referring to it directly.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -681,6 +678,30 @@ value will not be a valid X.509v3 signature that could be used in conjunction
...
+and in some cases the CA might not actually issue the corresponding certificate.
+Nevertheless, a precertificate's `signature` indicates the CA's binding intent
+to issue the corresponding certificate, which means that:
+
+* Misissuance of a precertificate is considered equivalent to misissuance of
+  the corresponding certificate. The CA should expect to be held to account,
+  even if the corresponding certificate has not actually been issued.
```
eranmes: add: "as clients have no way of verifying that the corresponding certificate has not been issued" or something along those lines?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -681,6 +678,30 @@ value will not be a valid X.509v3 signature that could be used in conjunction
...
+
+* Upon observing a precertificate, a client can reasonably presume that the
+  corresponding certificate has been issued. A client may wish to obtain status
+  information (e.g., by using the Online Certificate Status Protocol [RFC6960]
+  or by checking a Certificate Revocation List [RFC5280]) about a certificate
+  that is presumed to exist, especially if there is evidence or suspicion that
+  the corresponding precertificate was misissued.
+
+* TLS clients may have policies that require CAs to be able to revoke, and to
```
eranmes: I admit I find this paragraph a bit confusing - maybe drop "to be able to revoke, and" and just leave the bit about providing certificate status?

draft-ietf-trans-rfc6962-bis.md
robstradling: Addressed by https://github.com/robstradling/certificate-transparency-rfcs/commit/1a231c5a5300af1739842551fe373ec63ad4cd36.

draft-ietf-trans-rfc6962-bis.md
robstradling: I think it's unnecessary to add that text.

A precertificate contains information that the CA has digitally signed.  I think it's reasonable to say that a misissued precertificate, in and of itself, is something that the CA should be held to account for (even if it can be proven that the CA has not issued the corresponding certificate).

draft-ietf-trans-rfc6962-bis.md
robstradling: I think it's important to explicitly state "to be able to revoke", to echo the second of these two points from https://wiki.mozilla.org/CA/Required_or_Recommended_Practices#Precertificates:

> * A CA must provide OCSP services and responses in accordance with Mozilla policy for all certificates presumed to exist based on the presence of a Precertificate, even if the certificate does not actually exist
> * A CA must be able to revoke a certificate presumed to exist, if revocation of the certificate is required under Mozilla policy, even if the certificate does not actually exist.

I don't want CAs to think that they can refuse to revoke a precertificate just because the corresponding certificate hasn't actually been issued.

draft-ietf-trans-rfc6962-bis.md
eranmes: Acknowledged.

draft-ietf-trans-rfc6962-bis.md
eranmes: Acknowledged ,good point.

PR #317, opened 2021-03-08
Add trans:errors to URN sub-namespace

richsalz:
Also ask IANA to create the trans:error registry.

Also tweak some markdown syntax so my vim coloring isn't a screen
full of yellow :)

Fixes Alexey Melnikov's discuss items.


richsalz:
Closing this in favor of #318 which has the right branch info. :)\

PR #319, opened 2021-03-08
Add section on TLS server

richsalz:
Explain why server-side isn't discussed.  I.e., non-use of CT for TLS client certificates.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1827,6 +1827,14 @@ that are mandatory to implement for CT-using TLS clients (see
 if it did not include both the `transparency_info` and `status_request` TLS
 extensions in the ClientHello.
 
+## TLS Server {#no_tls_servers}
+
+This specification includes no description of what a TLS server should
+do if it wants to verify a received TLS client certificate.
+While this may be useful, it is not documented here because the greater
+security exposure is for clients to end up interacting with an illegitimate
+server. A future version could include such information.
+
 ## Monitor {#monitor}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1827,6 +1827,14 @@ that are mandatory to implement for CT-using TLS clients (see
 if it did not include both the `transparency_info` and `status_request` TLS
 extensions in the ClientHello.
 
+## TLS Server {#no_tls_servers}
```
robstradling, line 1830: I think this new section should be a subsection of the existing "TLS Server" section (search for "# TLS Servers {#tls_servers}" in the .md source).  Perhaps name this new section "TLS Client Authentication"?

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1827,6 +1827,14 @@ that are mandatory to implement for CT-using TLS clients (see
...
+
+This specification includes no description of what a TLS server should
+do if it wants to verify a received TLS client certificate.
```
robstradling, line 1833: I presume that "if it wants to verify" is intended to hint at the idea of using CT for TLS client certificates, but perhaps this could be stated more explicitly?
It might be worth pointing out that, in general, TLS client certificates are not expected to be submitted to CT logs, since "In this document, we only describe its use for public TLS server certificates" (see the Introduction).

richsalz:
pushed a new commit that moves the "client cert" sub-section, renames it as suggested, and adds another reason why not included.

PR #320, opened 2021-03-08
Clarify server status options

richsalz:
Three ways for a server to send CT data. Clarify when and how each is used.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1573,33 +1573,34 @@ Outputs:
 
 # TLS Servers {#tls_servers}
 
-CT-using TLS servers MUST use at least one of the three mechanisms listed below
+CT-using TLS servers MUST use at least one of the mechanisms described below
 to present one or more SCTs from one or more logs to each TLS client during full
 TLS handshakes, where each SCT corresponds to the server certificate. They
 SHOULD also present corresponding inclusion proofs and STHs.
 
-Three mechanisms are provided because they have different tradeoffs.
+The preferred mechanism is to use
+a TLS 1.3 extension (Section 4.2 of [RFC8446]) with type `transparency_info`
+(see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
+participate in CT without the cooperation of CAs, unlike the other two
+mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
 
-* A TLS extension (Section 4.2 of [RFC8446]) with type `transparency_info`
-  (see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
-  participate in CT without the cooperation of CAs, unlike the other two
-  mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
+If the server supports TLS version 1.2 or earlier, it may use
+an Online Certificate Status Protocol (OCSP) [RFC6960] response extension (see
+{{ocsp_transinfo_extension}}), where the OCSP response is provided in the
+`CertificateStatus` message, provided that the TLS client included the
+`status_request` extension in the (extended) `ClientHello` (Section 8 of
+[RFC6066]). This mechanism is popularly known as OCSP stapling.
+It also allows SCTs and inclusion
+proofs to be updated on the fly.
 
-* An Online Certificate Status Protocol (OCSP) [RFC6960] response extension (see
-  {{ocsp_transinfo_extension}}), where the OCSP response is provided in the
-  `CertificateStatus` message, provided that the TLS client included the
-  `status_request` extension in the (extended) `ClientHello` (Section 8 of
-  [RFC6066]). This mechanism, popularly known as OCSP stapling, is already
-  widely (but not universally) implemented. It also allows SCTs and inclusion
-  proofs to be updated on the fly.
-
-* An X509v3 certificate extension (see {{cert_transinfo_extension}}). This
-  mechanism allows the use of unmodified TLS servers, but the SCTs and inclusion
-  proofs cannot be updated on the fly. Since the logs from which the SCTs and
-  inclusion proofs originated won't necessarily be accepted by TLS clients for
-  the full lifetime of the certificate, there is a risk that TLS clients will
-  subsequently consider the certificate to be non-compliant and in need of
-  re-issuance.
+For off-line use,
+an X509v3 certificate extension (see {{cert_transinfo_extension}}). This
+mechanism allows the use of unmodified TLS servers, but the SCTs and inclusion
+proofs cannot be updated on the fly. Since the logs from which the SCTs and
+inclusion proofs originated won't necessarily be accepted by TLS clients for
+the full lifetime of the certificate, there is a risk that TLS clients will
+subsequently consider the certificate to be non-compliant and in need of
+re-issuance.
 
 ## Multiple SCTs {#multiple-scts}
 
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -1578,7 +1578,7 @@ to present one or more SCTs from one or more logs to each TLS client during full
 TLS handshakes, where each SCT corresponds to the server certificate. They
 SHOULD also present corresponding inclusion proofs and STHs.
 
-A server can provide SCT's using
+A server can provide SCTs using
 a TLS 1.3 extension (Section 4.2 of [RFC8446]) with type `transparency_info`
 (see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
 participate in CT without the cooperation of CAs, unlike the other two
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,34 @@ Outputs:
...
-  proofs to be updated on the fly.
-
-* An X509v3 certificate extension (see {{cert_transinfo_extension}}). This
-  mechanism allows the use of unmodified TLS servers, but the SCTs and inclusion
-  proofs cannot be updated on the fly. Since the logs from which the SCTs and
-  inclusion proofs originated won't necessarily be accepted by TLS clients for
-  the full lifetime of the certificate, there is a risk that TLS clients will
-  subsequently consider the certificate to be non-compliant and in need of
-  re-issuance.
+The preferred mechanism is to use
```
robstradling, line 1581: Although it was certainly the original intent in CT's early years, I think we would need to find out if there's still consensus to describe this mechanism as "preferred".

sleevi:
Thanks Rich for sending this.

Unfortunately, I don’t believe it is accurate or reflecting implementation experience to suggest the TLS 1.3 mechanism is preferred. The design of CT, and the ample implementation experience of both CAs and site operators, firmly supports embedded proofs as the preferred mechanism. This reduces server-side complexity and overhead for the general case (e.g. every TLS server having to manage the state of every UAs set of accepted Logs and required policies and appropriately ensure the necessary SCTs are sent).

It remains a quite nice option for CDNs who are staffed to ensure such things, just as they do with certificate and certificate chains, but I think it’s rather the opposite: the TLS and OCSP mechanisms are less preferred because they have high operational complexity, ongoing maintenance costs (since the SCTs must continue to comply with policy, as opposed to issuance-date policies that embedded SCTs have), and create a more fragile, less agile ecosystem.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,34 @@ Outputs:
...
+a TLS 1.3 extension (Section 4.2 of [RFC8446]) with type `transparency_info`
+(see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
+participate in CT without the cooperation of CAs, unlike the other two
+mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
+
+If the server supports TLS version 1.2 or earlier, it may use
```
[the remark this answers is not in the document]
richsalz: fixup commit pushed.

richsalz:
Push commit to address comments from @robstradling and @sleevi

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
+[RFC6960] response extension (see {{ocsp_transinfo_extension}}),
+popularly known as "OCSP stapling."
+For TLS
+1.3, the information is encoded as an extension in the `status_request`
+extension data; see Section 4.4.2.1 of [RFC8446] For TLS 1.2, the information
+is encoded as an extension in the `CertificateStatus` message; see Section 8
+of [RFC6066]).  This mechanism also
+allows SCTs and inclusion proofs to be updated on the fly.
+
+For off-line use,
```
AGWA, line 1597: "For off-line use" should be dropped as it implies that this mechanism is unsuitable or less preferred when the server is online, but I don't think this is supported by implementation experience.

richsalz: I removed "for off-line use" and tweaked the first sentence; additional commit pushed.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
-  proofs to be updated on the fly.
-
-* An X509v3 certificate extension (see {{cert_transinfo_extension}}). This
-  mechanism allows the use of unmodified TLS servers, but the SCTs and inclusion
-  proofs cannot be updated on the fly. Since the logs from which the SCTs and
-  inclusion proofs originated won't necessarily be accepted by TLS clients for
-  the full lifetime of the certificate, there is a risk that TLS clients will
-  subsequently consider the certificate to be non-compliant and in need of
-  re-issuance.
+A server can provide SCT's using
```
robstradling, line 1581: Nit: Drop the apostrophe.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
+(see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
+participate in CT without the cooperation of CAs, unlike the other two
+mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
+
+The server may also use an Online Certificate Status Protocol (OCSP)
+[RFC6960] response extension (see {{ocsp_transinfo_extension}}),
+popularly known as "OCSP stapling."
+For TLS
+1.3, the information is encoded as an extension in the `status_request`
+extension data; see Section 4.4.2.1 of [RFC8446] For TLS 1.2, the information
```
robstradling, line 1592: Missing full stop before "For TLS 1.2,"

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
+is encoded as an extension in the `CertificateStatus` message; see Section 8
+of [RFC6066]).  This mechanism also
```
robstradling, line 1594: ")" without a corresponding "(".

richsalz:
I pushed a commit that fixes the three typo's @robstradling found.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
+a TLS 1.3 extension (Section 4.2 of [RFC8446]) with type `transparency_info`
+(see {{tls_transinfo_extension}}). This mechanism allows TLS servers to
+participate in CT without the cooperation of CAs, unlike the other two
+mechanisms. It also allows SCTs and inclusion proofs to be updated on the fly.
+
+The server may also use an Online Certificate Status Protocol (OCSP)
+[RFC6960] response extension (see {{ocsp_transinfo_extension}}),
+popularly known as "OCSP stapling."
+For TLS
+1.3, the information is encoded as an extension in the `status_request`
```
sleevi, line 1591: Perhaps:

```suggestion
The server may also use an Online Certificate Status Protocol (OCSP)
[RFC6960] response extension (see {{ocsp_transinfo_extension}}),
providing the OCSP response as part of the TLS handshake. Providing
a response during a TLS handshake is popularly known as "OCSP stapling."
For TLS
1.3, the information is encoded as an extension in the `status_request`
```

The area of concern here is two-fold:
1. It reads that "OCSP stapling" is the act of using OCSP response extensions - rather than providing an OCSP response itself.
2. "The information is encoded" continues the ambiguity, by implying it's the extension data that's provided in these paths, rather than the response.

I'm not sold on the wording I suggested, so feel free to reject/editorialize at will, but I wanted to flag that ambiguity. In the original language, "This mechanism" naturally followed the "TLS client included the `status_request` extension", so was less ambiguous (but still slightly so)

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1573,33 +1573,35 @@ Outputs:
...
+allows SCTs and inclusion proofs to be updated on the fly.
+
+CT information can also be encoded as an extension in the X.509v3 certificate
+(see {{cert_transinfo_extension}}). This
+mechanism allows the use of unmodified TLS servers, but the SCTs and inclusion
+proofs cannot be updated on the fly. Since the logs from which the SCTs and
+inclusion proofs originated won't necessarily be accepted by TLS clients for
+the full lifetime of the certificate, there is a risk that TLS clients will
+subsequently consider the certificate to be non-compliant and in need of
+re-issuance.
```
sleevi, line 1604:
```suggestion
the full lifetime of the certificate, there is a risk that TLS clients may
subsequently consider the certificate to be non-compliant and in need of
re-issuance or the use of one of the other two methods for delivering CT
information.
```

I'm not sure the will/may is strictly necessary, but mostly it reflects the fact that we haven't (yet) had this situation play out, but it is a valid concern.

The other part is explicitly clarifying that the alternative methods may be sufficient to avoid reissuance. Reissuance is unambiguously the preferred path for the Web PKI (in the principle of GREASE, it's the well-greased path), but perhaps its useful to clarify that the other two methods exist in part to offer an alternative to that (since the state of the art in 2011/2012 wasn't as hopeful)

draft-ietf-trans-rfc6962-bis.md
richsalz: I took your suggestion, and also changed "This mechanism also..." at the end of the paragraph to be "Using stapling also ..."

PR #324, opened 2021-03-26
Mirja comments

richsalz:
This addresses the comments Mirja had during IESG review.  No changes were made for the following:
```
 7) sec 10.3: Couldn’t you use the TLS  SignatureScheme Registry directly (instead of forming an own registry)?

 8) sec 10.4: i Wonder if an RFC-required policy wouldn’t be more appropriate for the VersionedTransTypes registry?

 9) sec 10.6.1:I guess  the registration policy is FCFS? RFC 8126 recommend to always (clearly) name the registry.

And finally one higher level question mainly out of curiosity: was it considered to e.g. use json for all data structures? Is there a performance reason to not do that or wasn’t that even discussed?
```

Replies:
7 Using the TLS registry is not appropriate, the signatures are for different protocols even though the algorithm is the same
8 No.
9. OID's are FCFS and they're not really a registry.
Finally: signed json structures are complex and time-consuming, in addition to the verbosity over binary structures

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -869,7 +869,8 @@ Each log is identified by an OID, which is one of the log's parameters (see
 {{log_parameters}}) and which MUST NOT be used to identify any other log. A
 log's operator MUST either allocate the OID themselves or request an OID from
 the Log ID Registry (see {{log_id_registry}}).
-(Recall that OID's are do not require a central registration, although
+The only advantage of the registry is that the DER encoding can be small.
+(Recall that OID allocations do not require a central registration, although
 logs will most likely want to make themselves known to potential clients
 through out of band means.)
 Various data structures include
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -868,7 +868,11 @@ chain that the log used to verify the submission.
 Each log is identified by an OID, which is one of the log's parameters (see
 {{log_parameters}}) and which MUST NOT be used to identify any other log. A
 log's operator MUST either allocate the OID themselves or request an OID from
-the Log ID Registry (see {{log_id_registry}}). Various data structures include
+the Log ID Registry (see {{log_id_registry}}).
+(Recall that OID's are do not require a central registration, although
```
robstradling, line 872: Typo/Suggestion: s/Recall that OID's are do not/OID allocations do not/

richsalz: took the suggestion.  Also added a sentence. Fixup commit pushed.

PR #325, opened 2021-03-26
Alexey comments

richsalz:
This addresses Alexey's comments that haven't already been resolved.

There is one open question about base64, posted to the list as https://mailarchive.ietf.org/arch/msg/trans/a0MXL0pqVRKi8NB8199YoUQRlpo/  Once I get confirmation I will take this out of WIP; commit 737b95fe09868811f86be01975649143c608edb2 has the expected wording.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -52,6 +52,7 @@ normative:
   RFC8032:
   RFC8174:
   RFC8259:
+  RFC5246:
   RFC8446:
   HTML401:
     target: http://www.w3.org/TR/1999/REC-html401-19991224
@@ -1598,7 +1599,7 @@ providing the OCSP response as part of the TLS handshake. Providing
 a response during a TLS handshake is popularly known as "OCSP stapling."
 For TLS
 1.3, the information is encoded as an extension in the `status_request`
-extension data; see Section 4.4.2.1 of [RFC8446]. For TLS 1.2, the information
+extension data; see Section 4.4.2.1 of [RFC8446]. For TLS 1.2 ([RFC5246]), the information
 is encoded as an extension in the `CertificateStatus` message; see Section 8
 of [RFC6066].  Using stapling also
 allows SCTs and inclusion proofs to be updated on the fly.
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -331,7 +331,8 @@ used:
     1. Push `HASH(0x00 || entries[i])` to `stack`.
 
     2. Set `merge_count` to the lowest value (`0` included) such that `LSB(i >>
-       merge_count)` is not set. In other words, set `merge_count` to the number
+       merge_count)` is not set, where `LSB` means the least significant byte.
+       In other words, set `merge_count` to the number
        of consecutive `1`s found starting at the least significant bit of `i`.
 
     3. Repeat `merge_count` times:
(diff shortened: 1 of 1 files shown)


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -846,8 +858,8 @@ If a submission is accepted and an SCT issued, the accepting log MUST store the
 entire chain used for verification. This chain MUST include the certificate or
 precertificate itself, the zero or more intermediate CA certificates provided by
 the submitter, and the trust anchor used to verify the chain (even if it was
-omitted from the submission). The log MUST present this chain for auditing upon
-request (see {{get-entries}}). This prevents the CA from avoiding blame by
+omitted from the submission). The log MUST provide this chain for auditing upon
+request (see {{get-entries}}) so that the CA cannot from avoid blame by
```
AGWA, line 862: "cannot from avoid blame" -> "cannot avoid blame"

richsalz: fixed, thanks!

richsalz:
Confirmation from an expert (@ag) so taking this out of WIP.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1838,10 +1852,11 @@ It is up to a client's local policy to specify the quantity and form of
 evidence (SCTs, inclusion proofs or a combination) needed to achieve
 compliance and how to handle non-compliance.
 
-A TLS client can only evaluate compliance if it has given the TLS server the
+A TLS client can only determine compliance if it has given the TLS server the
```
robstradling, line 1855: The long-held WG consensus is that 6962-bis should not specify "client behaviour" (see slides 8 and 9 of https://datatracker.ietf.org/meeting/91/materials/slides-91-trans-4), except where necessary to participate in the CT protocol.  That's why we chose the verb "evaluate" rather than anything stronger.

Alexey suggested changing "evaluate" to "enforce", but I think that sounds a bit too strong.

Aren't "evaluate" and "determine" synonymous?  I think "evaluate" was just fine.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -1838,10 +1852,11 @@ It is up to a client's local policy to specify the quantity and form of
...
 opportunity to send SCTs and inclusion proofs by any of the three mechanisms
 that are mandatory to implement for CT-using TLS clients (see
-{{receiving_transitems}}). Therefore, a TLS client MUST NOT evaluate compliance
+{{receiving_transitems}}). Therefore, a TLS client MUST NOT make such a
+compliance determination
```
robstradling, line 1859: I think "MUST NOT evaluate compliance" was just fine.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -331,7 +340,8 @@ used:
     1. Push `HASH(0x00 || entries[i])` to `stack`.
 
     2. Set `merge_count` to the lowest value (`0` included) such that `LSB(i >>
-       merge_count)` is not set. In other words, set `merge_count` to the number
+       merge_count)` is not set, where `LSB` means the least significant byte.
```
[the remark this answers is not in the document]
richsalz: yes, should be bit.  fingers faster than brain.

draft-ietf-trans-rfc6962-bis.md
richsalz: I agree "enforce" is the wrong word to use.  To me, "evaluate" implies making a decision and taking action on it while "determine" is missing that action step. But I might be wrong, and will change it back.

draft-ietf-trans-rfc6962-bis.md
richsalz: reverted

PR #326, opened 2021-03-26
Proof paths can be zero-length

richsalz:
Ben Kaduk pointed out that the `ConsistencyProofDataV2` and `InclusionProofDataV2` had arrays that could not be empty. This was contradicted elsewhere in the document, so fix the `1..` to `0..` in both cases.

This should address the last part of Ben's DISCUSS.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -498,17 +498,17 @@ consistency proof:
 1. If `consistency_path` is an empty array, stop and fail the proof
    verification.
 
-1. If `first` is an exact power of 2, then prepend `first_hash` to the
+2. If `first` is an exact power of 2, then prepend `first_hash` to the
    `consistency_path` array.
 
-2. Set `fn` to `first - 1` and `sn` to `second - 1`.
+3. Set `fn` to `first - 1` and `sn` to `second - 1`.
 
-3. If `LSB(fn)` is set, then right-shift both `fn` and `sn` equally until
+4. If `LSB(fn)` is set, then right-shift both `fn` and `sn` equally until
    `LSB(fn)` is not set.
 
-4. Set both `fr` and `sr` to the first value in the `consistency_path` array.
+5. Set both `fr` and `sr` to the first value in the `consistency_path` array.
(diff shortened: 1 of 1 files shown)


robstradling:
The algorithm in Section 2.1.4.2 (Verifying Consistency between Two Tree Heads) requires at least 1 element in the consistency_path:
`4.  Set both "fr" and "sr" to the first value in the "consistency_path" array.`

This section also says `where "0 < first < second"`, which means that at least 1 element must be present for the consistency_path to be valid (because first != second).  However, we can't assume that the consistency_path is valid until it has been validated.

So I suggest adding a new first step to the algorithm, with the following description:
`If consistency_path is an empty array, stop the iteration and fail the proof verification.`

WDYT?

richsalz:
> If consistency_path is an empty array, stop the iteration and fail the proof verification.

looks good; updated commit pushed.

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -495,6 +495,9 @@ received a consistency proof between the two (e.g., in a `TransItem` of type
 `consistency_proof_v2`), the following algorithm may be used to verify the
 consistency proof:
 
+1. If `consistency_path` is an empty array, stop and fail the proof
+   verification.
+
 1. If `first` is an exact power of 2, then prepend `first_hash` to the
```
robstradling, line 501: Now there are two step 1s.  Please renumber the list of steps.

richsalz: markdown renumbers numbered lists so this is not necessary.  to avoid confusion I renumbered them all to be "1." :)

robstradling: I see 1, 1, 2, 3, 4, 5, 6 at the moment.

robstradling: I would much prefer the numbering in the .md source to be correct, so that anyone who reads the .md source in the future does not get confused.

richsalz: fine, changed and pushed.

PR #338, opened 2021-07-14
Be consistent in SCT/STH acronym use

richsalz:
Make sure it's spelled out the first time it's used.
Use acronym in section title except for the definition and include
the acronym then.


paulwouters:
looks good to me, and I think this can be pulled in during RFC-Editor work or it can be merged in a new draft after our IETF 111 meeting if anything else comes up. Thanks Rich !

PR #340, opened 2021-07-29
Fix punctuation to aovid run-on sentence

richsalz:


draft-ietf-trans-rfc6962-bis.md
```diff
@@ -852,7 +852,7 @@ vendors.
 ### Minimum Acceptance Criteria {#minimum_criteria}
 
 To ensure that logged certificates and precertificates are attributable to an
-accepted trust anchor, and to set clear expectations for what monitors would
+accepted trust anchorand to set clear expectations for what monitors would
```
robstradling, line 855: I think Erik just wanted us to remove the first "and".
```suggestion
accepted trust anchor, to set clear expectations for what monitors would
```

robstradling: Erik's comment was:
[S4.2.1] [nit]

* The first sentence reads like a bit of a run-on sentence to me. Perhaps:

  "To A, and to B, and to C, the log MUST..." ->
  "To A, to B, and to C, the log MUST..."

robstradling (changes requested):

robstradling (approved):

PR #341, opened 2021-07-29
Add ref to Private Enterprise Number page.

richsalz:
Per discussion with Martin re his IESG comments.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -911,6 +911,9 @@ Each log is identified by an OID, which is one of the log's parameters (see
 {{log_parameters}}) and which MUST NOT be used to identify any other log. A
 log's operator MUST either allocate the OID themselves or request an OID from
 the Log ID registry (see {{log_id_registry}}).
+One way to get an OID arc, from which OIDs can be allocated, is to request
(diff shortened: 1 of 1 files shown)


richsalz:
rebased and pushed.  it really is a trivial thing :)

draft-ietf-trans-rfc6962-bis.md
```diff
@@ -911,6 +911,9 @@ Each log is identified by an OID, which is one of the log's parameters (see
 {{log_parameters}}) and which MUST NOT be used to identify any other log. A
 log's operator MUST either allocate the OID themselves or request an OID from
 the Log ID registry (see {{log_id_registry}}).
+One way to get an OID arc, from which OIDs can be created, is to request
```
robstradling, line 914: Nit: We use "allocate" in the previous paragraph, so let's use it here too.
```suggestion
One way to get an OID arc, from which OIDs can be allocated, is to request
```

robstradling (changes requested):

draft-ietf-trans-rfc6962-bis.md
richsalz: created->allocated is good; updated commit pushed.

PR #346, opened 2021-08-16
Renumber VersionTransType entries to actually ensure that v2 structures are always distinguishable from v1 SCTs.

robstradling:
https://trac.ietf.org/trac/trans/ticket/157 made a basic error regarding the range of uint16 values for which the first byte (in network byte order) is zero.  Per discussion in https://mailarchive.ietf.org/arch/msg/trans/O3oUjxOYr0iBVvjoBfrL5V15r08/, this PR proposes to address this error by changing the reserved ranges and renumbering the initial registry entries accordingly.

### DIFF
--- a/draft-ietf-trans-rfc6962-bis.md
+++ b/draft-ietf-trans-rfc6962-bis.md
@@ -942,12 +942,16 @@ that the type and version of each one is identified in a common fashion:
 
 ~~~~~~~~~~~
     enum {
-        reserved(0),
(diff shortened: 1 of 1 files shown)


richsalz:
Still unconvinced of the need for this, but sure if this is more safe, let's do it.

eranmes (approved):
looks good to me, thanks for fixing this Rob.

